PR #1958, opened 2026-08-05
fix(admin-ui): lazy OIDC discovery — the console boots and serves Basic login while the issuer is unreachable

### LINKED ISSUE #1950: fix(admin-ui): eager OIDC discovery at boot crash-loops the console on an unreachable issuer — degrade to Basic login instead
The admin console performs eager OIDC discovery at boot and crashes when the issuer is unreachable, where the CDR fails lazily per request. Found during the #1948 quickstart investigation (2026-08-05).

Facts:

- `app/ferroehr-admin-ui/src/main.rs` runs `oidc::discover(...).await?` at startup when `auth.oidc.enabled = true` — an unreachable/slow issuer aborts the whole console instead of degrading to Basic login with OIDC marked unavailable.
- The CDR's equivalent (`app/ferroehr-rest/.../authn/jwt.rs`) is lazy: construction is inert, discovery happens on the first bearer token, failure is a 401.
- In a composed deployment this creates a startup-order coupling: the console container crash-loops until Keycloak's realm import finishes, even though its Basic login path needs nothing from the IdP.

Wanted: the console boots with OIDC configured-but-unreachable, serves Basic login (when enabled), retries discovery in the background or on first OIDC login attempt, and surfaces the IdP outage in the login UI instead of dying.

## Acceptance criteria

- [ ] With `auth.oidc.enabled = true` and an unreachable issuer, the console boots and serves `/login` with the Basic form (when Basic is enabled).
- [ ] OIDC login attempts during the outage fail with a clear in-UI message; discovery is retried (background or on-demand) and recovers without a restart.
- [ ] A test pins the boot-with-unreachable-issuer behaviour.

rubentalstra:
The console ran OIDC discovery at boot and exited on failure, so an unreachable issuer crash-looped the container — while the CDR's equivalent path (`ferroehr-rest` `authn/jwt.rs`) has always been lazy. In a composed deployment that made the console's startup depend on Keycloak's realm import, even though its Basic login path needs nothing from the IdP. Discovery is now lazy, cached, and self-retrying.

Closes #1950

## What changed

**`src/oidc.rs` — lazy, cached, self-retrying discovery.** `oidc::discover()` is replaced by `OidcState::new(&OidcConfig)`, which does **no network I/O**: it builds the openidconnect reqwest client (keeping `redirect::Policy::none()` and the split-horizon `resolve` override) and parses the issuer URL. Everything it checks is LOCAL config — a malformed `resolve`, an unparseable address, a bad issuer URL — so those stay fatal at boot, as before. A new `discovered: tokio::sync::Mutex<Option<CoreProviderMetadata>>` backs `async fn metadata()`, which returns the cached document or runs `CoreProviderMetadata::discover_async`. A failure caches **nothing**, so the next OIDC sign-in retries — recovery without a restart; the tokio mutex collapses concurrent first sign-ins into one discovery instead of a thundering herd. `client()` became synchronous, taking the metadata as an argument, so a provider outage and a malformed `public_base_url` stay distinct failure classes. `prime_in_background(Arc<OidcState>)` runs one spawned, non-fatal warm-up so an IdP that is down at startup is named loudly in the boot log rather than only surfacing on someone's first sign-in.

**In-UI outage reporting.** A new `idp_unavailable(issuer, err)` helper logs the raw diagnostic (operators need it; users cannot act on it) and redirects to `/login?error=<urlencoding::encode(msg)>` with actionable copy: *"OIDC sign-in is unavailable: the identity provider at {issuer} could not be reached. Try again shortly, or sign in with a username and password."* Both `login` and `callback` route their discovery failures through it; every other flow error keeps its existing handling. `pages/login.rs` derives `?error=` from the `use_query_map` it already holds and renders it in the **same** thaw MessageBar the bad-credential path uses — a live action error wins over a stale URL param. Deterministic from the URL, so hydration-safe.

**New typed error variant.** `AdminUiError::IdpUnavailable { issuer, message }` — a distinct variant rather than a stringified `Internal`, because the console's answer to it differs from a CDR outage (offer the other login mode). The compiler's exhaustiveness check surfaced `feedback.rs::write_failure_copy`, which gained a matching arm.

**New `src/server.rs` (ssr-gated).** `pub fn router(AppState, LeptosOptions) -> axum::Router` moves the route table and session layer out of `main.rs`, so the boot path is importable from `tests/` (the Book's thin-`main`-over-`lib` split, ch12.3). `main.rs` keeps tracing init, the unchanged `healthcheck` subcommand, config load, state assembly, the background prime, bind and serve.

## Tests

`tests/it/boot_oidc_outage.rs`, registered as a `#[cfg(feature = "ssr")] mod` (the featureless lane cfg's it out), pins both halves of the contract:

- **`login_is_served_with_the_basic_form_while_the_issuer_is_unreachable`** — AppState with `issuer = https://idp.invalid/realms/console` and `cdr.base_url = http://cdr.invalid:8080` (RFC 2606 §2 reserved TLD: fails fast, and a developer's own CDR on :8080 cannot make it flaky), `server::router` served on an ephemeral port, `GET /login` → 200 containing `id="login-username"`. Under the old code this test could not even be written — `discover()` would have failed before AppState existed.
- **`discovery_retries_after_an_outage_and_caches_once_it_succeeds`** — a local stand-in provider answers `503` once, then serves a valid discovery document. The first `metadata()` yields `IdpUnavailable`; the **same** `OidcState` then discovers successfully (recovery without a restart); a third call adds no round trip, and the request counter must be exactly `2`. Two-sided: a cached failure fails the second assertion, an uncached success fails the third.

## Gates

| Gate | Result |
|---|---|
| `cargo clippy -p ferroehr-admin-ui --all-targets --features ssr -- -D warnings` | green |
| `cargo clippy -p ferroehr-admin-ui --target wasm32-unknown-unknown --features hydrate -- -D warnings` | green |
| `cargo nextest run -p ferroehr-admin-ui --features ssr` | 272/272 passed |
| `cargo nextest run -p ferroehr-admin-ui` (featureless) | 218/218 passed |
| `cargo fmt --all --check` · `leptosfmt --check` | clean |
| `scripts/check-comment-style.sh` (8 touched files) | OK |

`cargo leptos build` was **not** run (stopped at the owner's request to push immediately); CI's build and `ui-e2e` jobs gate the merge. The change touches the login journey that `scripts/ui-e2e.sh` covers — not run locally either.

## Docs

`CHANGELOG.md` `[Unreleased]` → `### Fixed` (end-user voice) and one paragraph in `website/book/src/admin-ui/index.md`. **No configuration keys change.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ workflow refuses a tag that has no matching section here.
 
 - The `openehr-rm` and `openehr-base` crates now ALSO emit the latest RELEASED specification generations beside the development pins — `openehr_rm::v1_1` (RM 1.1.0, resolving against BASE 1.2.0 per its own BMM `includes`) and `openehr_base::v1_2` (BASE 1.2.0) — each a complete peer: full type model, canonical-JSON codecs, RM attribute model, invariant cores, and validation behaviour. The served wire is unchanged (the current generations stay `v1_2`/`v1_3`); runtime selection between generations arrives with the `spec_profile` configuration key.
 
+### Fixed
+
+- The admin console no longer refuses to start when OIDC is enabled but the identity provider is unreachable. It used to run discovery at startup and exit on failure, so in a composed deployment the console crash-looped until the identity provider finished importing its realm — even though its username/password sign-in needs nothing from that provider. The console now starts immediately and serves the sign-in page: username/password login works throughout the outage, the OIDC button reports "OIDC sign-in is unavailable: the identity provider at … could not be reached" instead of failing blankly, and the outage is named in the startup log. Discovery is retried on the next OIDC sign-in, so OIDC starts working by itself once the provider answers — no console restart needed. Genuinely broken OIDC *configuration* (a malformed issuer URL, redirect URL, or `resolve` override) still stops startup, as before.
+
 ### Removed
 
 - The generated `openehr-*` spec crates no longer carry a crate-level `SPEC_VERSION` constant: a multi-generation crate has no single implemented spec version, and a fixed crate-root pin would contradict a configured non-current generation. The ONLY pin authority is the emitted `Generation` enum (per-variant `const fn spec_version()`; the derived `Default` variant is the current generation) — the generation modules carry no version constant either; the hand-written single-spec crates (`openehr-its`, `openehr-query`, `openehr-adl`) keep their literal constant.
--- a/app/ferroehr-admin-ui/src/error.rs
+++ b/app/ferroehr-admin-ui/src/error.rs
@@ -28,6 +28,17 @@ pub enum AdminUiError {
     /// The CDR was unreachable (connect/timeout/transport).
     #[error("CDR unreachable: {0}")]
     CdrUnreachable(String),
+    /// The OIDC identity provider could not be reached, so the OIDC sign-in
+    /// path is unavailable while Basic login keeps working. Distinct from
+    /// [`Self::CdrUnreachable`]: a different dependency is down, and the
+    /// console's answer is different (offer the other login mode).
+    #[error("identity provider at {issuer} unavailable: {message}")]
+    IdpUnavailable {
+        /// The configured `auth.oidc.issuer`.
+        issuer: String,
+        /// The discovery diagnostic (transport, status, or malformed document).
+        message: String,
+    },
     /// Bad user input (login form, AQL text, template upload) with the
     /// message to render inline.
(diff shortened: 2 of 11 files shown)


rubentalstra:
ON HOLD (owner decision 2026-08-05): not merging yet — concern that degrading to Basic login when the IdP is unreachable weakens the console's security posture. Held open pending an owner security review of the degrade-vs-refuse trade-off.

PR #2174, opened 2026-08-07
v3.17.4 batch: signing rotation, audit tamper detection, DDL-free runtime role, CSP nonce, bootable chart values

### LINKED ISSUE #2049: security: the shipped deployment role can run DDL at runtime, because migrations run at boot
Found while proving #2025's least-privilege claim, and it is the honest version of a claim the docs stated too flatly.

The layered roles exist and work: a role that is a member of **`ferroehr_app` only** was verified to hold none of `rolsuper`/`rolbypassrls`/`rolcreaterole`/`rolcreatedb`/`rolreplication`, and to be refused `CREATE TABLE`, `DROP TABLE`, `ALTER TABLE`, `CREATE SCHEMA`, `CREATE INDEX`, `TRUNCATE`, `CREATE ROLE` and `ALTER TABLE … DISABLE ROW LEVEL SECURITY`, while `SELECT` still works.

**But that posture is not the one the shipped deployment runs.** `app/ferroehr-server/src/lib.rs` calls `db::run_migrations` unconditionally at boot, and migrations are DDL — so the connecting role must be able to perform DDL. The quickstart's role accordingly owns the database and is a member of **both** `ferroehr_migrator` and `ferroehr_app`. Convenient, and correct for an evaluation; not least privilege for a clinical deployment.

The choice is real, not a bug: self-migrating on boot is what makes `docker compose up` work, and an operator who wants the tighter posture needs a way to opt in.

## Acceptance criteria
- [ ] A way to boot with migrations **not** applied by the server (so the DSN can be a `ferroehr_app`-only role), with the schema-version check still refusing to serve against an unmigrated or stale database
- [ ] The Helm chart able to run migrations as a separate step (a job or init container) under the migrator role, with the server under the app role
- [ ] The book stating both postures plainly: what the quickstart does and why, and what a least-privilege deployment does instead
- [ ] The verification kept as a test, so the `ferroehr_app` refusals cannot silently regress

Documentation was corrected in-branch under #2025 (`website/book/src/operations.md` and `installation/compose.md` now say the connecting role belongs to both groups); this issue is the capability.

### LINKED ISSUE #2051: security(admin-ui): replace the console CSP inline allowance with a per-request nonce
The console now carries the browser security-header set with a CSP (#2016), and that policy contains `'unsafe-inline'` for both `script-src` and `style-src`. This issue is the honest remainder: making the script half strict.

Why it is there. Leptos emits its hydration bootstrap as an **inline** `<script>`, and `thaw` injects generated `<style>` elements — so an inline-free policy blocks hydration and the console renders but never comes alive. `'wasm-unsafe-eval'` is already the narrow form (WebAssembly compilation without `eval` of JavaScript), and the policy still forecloses external script origins, object embedding, framing and cross-origin connections, which is most of the realistic surface for a same-origin SSR console. But an inline allowance means an injected inline script would run.

Why it is separate rather than done. The strict form is a per-request nonce, and Leptos supports it — `leptos::nonce::{Nonce, provide_nonce, use_nonce}` behind the crate's `nonce` feature, with `HydrationScripts` stamping `nonce=` on its inline script when a `Nonce` is in context. The plumbing is available too: `leptos_axum` puts the request `Parts` into the render context, so a middleware can mint `Nonce::new()`, carry it in request extensions, and set the CSP header with the same value that the shell provides into context.

What makes it a separate step is verification, not difficulty. A nonce that does not reach the emitted script produces a console that looks fine server-side and is dead in the browser — so this needs a real hydration check (the composed stack plus a browser), not a plausible-looking header and a green clippy.

## Acceptance criteria
- [ ] The `nonce` feature enabled on the server target, a nonce minted per request and carried into the render context
- [ ] `script-src` reduced to `'self' 'wasm-unsafe-eval' 'nonce-…'` with no `'unsafe-inline'`
- [ ] Hydration verified in a browser against the composed stack — the console loads, navigates, and a server function round-trips
- [ ] `style-src`: either nonced the same way, or the `'unsafe-inline'` retained with `thaw`'s lack of nonce support recorded as the reason
- [ ] A test asserting the emitted script's nonce matches the header's

### LINKED ISSUE #2059: security(audit): the audit trail is append-only by convention only — no tamper detection
Found by checking the surface against the OWASP **Logging** Cheat Sheet, which the earlier audit never enumerated. Controls: §Log Integrity — *"Build in tamper detection so you know if a record has been modified or deleted"*, *"store or copy log data to read-only media as soon as possible"*, *"separate database account with restrictive permissions."*

The local Audit Record Repository (`audit_event`) is **append-only by convention only**: the property is asserted in a migration comment and enforced by nothing — no hash chain, no per-record signature, no `REVOKE UPDATE/DELETE`, no trigger — and the application connects with the same role that owns the schema. The off-box copies (RFC 5424 syslog, ITI-20 ATX:FHIR feed) are the cheat sheet's "copy to separate media" control, and they are **off by default**.

So in the default deployment, anything holding the application's database credentials can silently rewrite or delete the record of who accessed which patient.

The asymmetry is what makes this worth fixing rather than noting: **this project already ships the primitive.** `ferroehr::signing` integrity-protects clinical VERSION content. The clinical data is signed and the record of who touched it is not — and in a regulated setting the audit trail is the accountability artifact, so its integrity is the load-bearing one.

Related but distinct: #2049 covers the DDL-privilege half of the connecting role. This is about the records themselves.

## Acceptance criteria
- [ ] Tamper detection on `audit_event` — a hash chain over the preceding record, a per-record signature reusing the existing signer, or both — with the choice justified against the ATNA/DICOM PS3.15 requirements and the write-path cost
- [ ] A verification path an operator can run: given the repository, report whether any record was modified or removed, and which
- [ ] `REVOKE UPDATE, DELETE` on the table for the application role, with the least-privilege deployment posture stated (see #2049)
- [ ] A test that forces a modification and proves detection, rather than asserting the property in a comment

### LINKED ISSUE #2122: feat(signing): rotating a PGP signing key makes every previously-signed version fail verification
Found while documenting key rotation for the Kubernetes hardening chapter (#2099). The question was "how do previously-signed versions remain verifiable across a rotation", and the answer is that they do not.

Two facts from the verification path, read rather than assumed:

- **A stored signature carries no key identifier.** `app/ferroehr/src/versioning/signature/verify.rs` records signature *text* and nothing else, so verification classifies the signature **by format** and then checks it against whatever key is configured now.
- **There is exactly one key, and no keyring.** `signing.key_path: Option<PathBuf>` loads a single `SignedSecretKey`. There is no second key, no key set, no per-signature lookup.

So after a rotation every version signed with the previous key verifies as `pgp_invalid`. With the default `verify_on_read: strict` that is an integrity failure raised while **reading historical clinical data** — a `5xx` on records that are perfectly intact. And the signatures cannot be re-issued: a `VERSION` signature is an immutable committed fact (RM common `master06` §Digital Signature), so re-signing would mean rewriting change-controlled history.

The workarounds a deployment has today are all reductions rather than solutions, and the hardening chapter states them as such: treat the signing key as long-lived and never rotate it; move to `verify_on_read: warn` across a rotation as a recorded reduction in guarantee; or use `digest` mode, which has no key and therefore no rotation problem but also no signer identity.

**Why this is a product gap rather than a documentation gap:** a deployment with a governance requirement for periodic signing-key rotation — which is an ordinary requirement in this sector — cannot satisfy it in `pgp` mode while keeping read-time verification of history. The capability that is missing is verification against a *set* of trusted keys, with the retired ones kept for verification only.

## Acceptance criteria
- [ ] A decision on the mechanism, with its reasoning: a verification keyring (retired keys retained, verify-only), a key identifier recorded alongside each signature so verification can select, or a documented decision that rotation is out of scope for `pgp` mode
- [ ] If a keyring: retired keys are never usable for signing, and a signature that matches none of them still fails — the point is to keep history verifiable, not to make verification permissive
- [ ] A test that signs with key A, rotates to key B, and reads the key-A version back under `verify_on_read: strict` without an integrity failure
- [ ] The hardening chapter's rotation section replaced with the real procedure, since it currently documents the workarounds

### LINKED ISSUE #2127: security(authn): AuthError::InvalidToken flattens the jsonwebtoken error, so expired and malformed are indistinguishable
Found while completing #2115; same defect class, outside that issue's named scope.

`app/ferroehr-rest/src/extensions/access/authn/jwt.rs` (two sites) builds `AuthError::InvalidToken(e.to_string())` from `jsonwebtoken`'s error. The crate distinguishes `ExpiredSignature`, `InvalidAudience`, `InvalidIssuer`, `InvalidSignature` and malformed input as separate variants; all of them arrive here as one string.

Two consequences, and the second is the one that matters:

- Nothing can branch on why a token was refused without matching a substring against another crate's message text — which changes with its version and is not part of its API.
- **An operator cannot tell a clock problem from an attack.** A burst of `ExpiredSignature` is usually clock skew or a token lifetime that is too short; a burst of `InvalidSignature` is someone presenting tokens this server was never meant to accept. Today both are the same log line, so the second hides in the first.

Wire behaviour must not change: every one of these is a `401` with a `WWW-Authenticate` challenge, and the response body must keep disclosing nothing about *why* — telling an unauthenticated caller that their signature was wrong rather than their audience is an oracle. The distinction belongs in the log and in the error type, never in the response.

## Acceptance criteria
- [x] `AuthError::InvalidToken` carries the underlying `jsonwebtoken::errors::Error` as a `#[source]`, or a typed reason enum mapped from it
- [ ] The `401` body and its `WWW-Authenticate` challenge are byte-identical for every refusal kind — asserted, since this is the property that makes the change safe
- [x] The refusal reason reaches the log as a structured field, so expired-versus-invalid is countable rather than greppable

### LINKED ISSUE #2154: test(config): template_mentions_every_section checks SECTIONS, so a missing key is invisible
Found by the new `docs-claims` guard (#1984), which resolves every documented `FERROEHR__…` form against `app/ferroehr/assets/ferroehr.default.toml` and reported one it could not find. The docs were right; the template was incomplete.

`auth.oidc.hmac_secret_file` and `auth.oidc.jwks_json_file` are real fields on `OidcConfig` (`app/ferroehr/src/config/auth.rs`), and both appeared in the shipped template only inside a trailing comment on a *different* key's line:

```toml
#? hmac_secret = "..."            # or hmac_secret_file = "/run/secrets/hmac"
#? jwks_json = "{...}"            # or jwks_json_file = "/etc/ferroehr/jwks.json"
```

So `ferroehr config default` — which the template's own header calls "the complete FerroEHR server configuration", with every key "shows its built-in default" — did not teach them as keys. That is the worst place for the gap to be: both are the file-indirection form of a *secret*, which is exactly what an operator goes looking for. Fixed in the #1984 branch by giving each its own `#?` line.

**The template has two tests and neither could catch it.** `template_parses_to_default` asserts the template round-trips to `FerroEhrConfig::default()` — a commented-out key is absent from both sides, so it passes. `template_mentions_every_section` iterates `alias::SECTIONS` and asserts each SECTION appears. A section is present the moment one of its keys is; the other keys in it are unchecked. Both tests pass on a template missing an arbitrary number of keys.

The check that would have caught it is the field-level version of the second test: every field of every config struct appears in the template as a key. That is not derivable by grep — it needs the struct definitions — but it is exactly what serde already walks, so serializing `FerroEhrConfig::default()` to TOML and asserting each emitted key path appears in the template (as a live line or a `#?` reference line) is a tractable shape. `Option::None` fields are the interesting case, since they serialize away — those are precisely the `#?` ones, so the test likely needs the field list from a different angle than a default-value serialization.

Filed rather than fixed because `app/ferroehr/src/config/` is being refactored under #2124 in the same tree; this belongs in a change that owns that file.

## Acceptance criteria
- [ ] A test that fails when a configuration field has no key line in `ferroehr.default.toml`, covering optional/secret/derived fields (the `#?` class) as well as ones with defaults
- [ ] Its failure message names the missing field path, so the fix is obvious without reading the test
- [ ] Mutation-proven: delete one key line from the template and confirm the test fails naming that key
- [ ] `template_mentions_every_section` kept or subsumed deliberately — if the new test makes it redundant, say so rather than leaving two tests that look like they check the same thing

### LINKED ISSUE #2159: test(deploy): the chart values files we ship do not boot
Found while fixing #2144. Every committed file under `deploy/helm/ci/` renders cleanly and **none of them produces a configuration the server accepts**:

- `default-values.yaml` — `auth.enabled = true` with no mechanism configured. `app/ferroehr/src/config/auth.rs:140` refuses it: *"auth.enabled = true but no authentication mechanism is configured"*. `values.yaml:299-306` documents that this refuses to boot, so the file is one item short of the "minimal realistic production overlay" its own comment claims to be.
- `basic-auth-values.yaml:17` — the placeholder hash fails the PHC check: *"not a PHC hash string: output size too short"*.
- `all-features-values.yaml:9` — a 15-byte HMAC against the RFC 8725 §3.5 32-byte floor, plus a missing `smart.public_base_url`.

This is a **release blocker**: the tag lane's acceptance step starts the `appVersion` image against the rendered default, so the next cut fails regardless of which image it selects.

The reason it survived is that `deploy/helm/validate.sh` renders and lints; nothing ever boots a container against the render. The gate that would have caught it is the one the publish lane runs — but only on a tag.

## Acceptance criteria
- [ ] Each values file produces a configuration the current image accepts, verified by running `config check` in the image against the render
- [ ] The values remain honest about their purpose: the "minimal production overlay" stays minimal, and a credential added only to satisfy boot is clearly a fixture
- [ ] A CI lane starts every committed values file against the image, so this cannot recur — not only on a tag
- [ ] `deploy/helm/validate.sh` says in its output which properties it does NOT check, so nobody reads a green render as a working deployment

### LINKED ISSUE #2172: feat(signing): sign with a rotatable signing subkey, not the primary key
Split out of #2122, which establishes the mechanism and the prior art.

We sign with the primary key:

```rust
// app/ferroehr/src/versioning/signature/key.rs
let sig = DetachedSignature::sign_binary_data(OsRng, &self.secret.primary_key, …);
```

RFC 9580 §5.2.1.8 and every practitioner guide say the opposite: sign with a **signing subkey**, keep the primary key for identity and certification, and rotate the subkey. That is what makes rotation cheap — the certificate keeps the retired subkey, so history verifies against the same configured key with no operator ceremony and no keyring entry.

#2122 fixes the verification half (a signature by any bound subkey now verifies). This is the signing half. Until it lands, our own rotations must go through the retired-keyring path, which works but is the heavier of the two mechanisms and the one meant for a genuine certificate change rather than routine rotation.

## What makes it more than a one-liner

The subkey has to be selected by **capability**, not by position. An OpenPGP certificate typically carries an encryption subkey and may carry several subkeys with different flags; signing with the wrong one produces a key-usage violation that a strict verifier should reject. `pgp` 0.20 does not surface a convenient `is_signing_key()` on `SignedSecretSubKey`, so this needs reading the binding signature's Key Flags subpacket (RFC 9580 §5.2.3.29, flag `0x02` "This key may be used to sign data") and choosing among the subkeys that carry it.

Behaviour must not change for a key that has no signing subkey: fall back to the primary key exactly as today, so existing deployments are untouched by an upgrade.

## Acceptance criteria
- [ ] A signing-capable subkey is selected by its Key Flags, never by position, and a certificate with no such subkey still signs with the primary key
- [ ] A test that signs with a subkey-bearing certificate and verifies the result, confirming the signature is attributable to the subkey and not the primary key
- [ ] A rotation test: sign with subkey A, add subkey B and retire A within the same certificate, and read the A-signed version back under `verify_on_read: strict` — with no `retired_key_paths` entry, since that is the point
- [ ] The hardening chapter's rotation procedure states subkey rotation as the normal path and the retired keyring as the exception

rubentalstra:
Closes #2122. Closes #2172. Closes #2154. Closes #2127. Closes #2049. Closes #2059. Closes #2051. Closes #2159.

Eight per-issue commits. Two of them started as one issue and turned into something larger once the prior art was read.

## Signing rotation (#2122 + #2172) — the research changed the answer

Rotating a PGP key made every previously-signed version verify as `pgp_invalid` — a 5xx while reading intact clinical records. Looking at how this is already solved found a defect underneath the issue: **OpenPGP provides the mechanism and we could not use it.** RFC 9580 §10.1 defines a transferable public key as a primary key *plus its subkeys*, and rotation means rotating the signing **subkey**, keeping the primary key's accumulated trust. Both ends were unavailable — `pgp` 0.20's `VerifyingKey for SignedPublicKey` consults `self.primary_key` alone and never walks `public_subkeys`, and we signed with the primary key.

So two layers now:

| layer | covers | mechanism |
|---|---|---|
| **subkey signing + full-certificate verification** | ordinary rotation | selected by RFC 9580 §5.2.3.29 key flag `0x02` — by **capability**, never position; verification tries the primary key then every subkey |
| **`signing.retired_key_paths`** | compromise, org change, migrating signers | verify-only **public** keys — no secret is loaded, so a retired certificate structurally *cannot* sign again |

The second is Debian's model for its own archive ([keys](https://ftp-master.debian.org/keys.html)), which is the closest analogue: a long-lived archive of signed artefacts that must stay verifiable across key changes.

Four tests pinning properties, not implementation. The one that earns its place is `a_certificate_with_a_signing_subkey_signs_and_verifies_through_it`: its primary key is `can_sign(false)`, so signing fails the fail-closed boot check if the subkey is not selected, **and** verification fails if subkeys are not walked — one test, both layers. The rotation test asserts both halves: verifies **with** the retained certificate and fails **without** it, so it proves the mechanism rather than its mere presence.

**Not solved, stated rather than implied:** key expiry and revocation. Nothing in a detached signature proves *when* it was made, so a verifier cannot separate "signed while valid" from "signed after expiry". That needs RFC 3161 timestamps — [Sigstore's](https://docs.sigstore.dev/about/security/) approach — and is not implemented.

The hardening chapter's rotation section documented three *workarounds* because no procedure existed; it now leads with subkey rotation.

## Config template completeness (#2154)

`template_parses_to_default` passes on a template missing any number of keys, and `template_mentions_every_section` checks sections — present the moment one key is. Six keys had shipped undiscoverable to `ferroehr config default`, **five of them the file-indirection form of a secret**, mentioned only inside another key's trailing comment.

The new walk over the serialized default is total for concrete fields with no hand-maintained list. Optional fields serialize away, so a second test takes its field list **from serde itself** (`deny_unknown_fields` answers with `expected one of …`) — with a guard that fails loudly if that message format changes, because a test silently checking nothing is worse than none. Mutation-proven both ways.

## Authentication (#2127) — and a leak found on the way

`TokenRejection` gives 20 variants with the `jsonwebtoken` error as a real `source()` — `#[error("{0}")]`, not `transparent`, which forwards `source()` and would erase the inner error. RFC 6750 §3.1 was read first-hand: `invalid_token` covers "expired, revoked, malformed" — one wire code for all — so the wire is correct and unchanged.

**The 401 body was an oracle.** It rendered the rejection to an unauthenticated caller (`invalid bearer token: ExpiredSignature` vs `… InvalidSignature`), separating "a real token, just stale" from "my forgery was caught". Nothing pinned it — no CNF case, no test, verified before changing — so the body is now generic and the reason moves to the log.

## Database trustworthiness (#2049 + #2059)

The runtime role held DDL rights because migrations ran at boot; `db.migrate = verify` never issues DDL and refuses to serve against a database that is not exactly this build's schema. The audit trail was append-only *by convention*; it now carries a hash chain.

**Eight tests against a real PostgreSQL**, asserting properties: a modified record detected **and named**; a deletion caught mid-chain **and at the end** (the case a naive chain misses); the application role refused DDL; `verify` refusing a database whose migration *text* diverged; and the **batched** write path chained too — the path that would otherwise be the hole.

## Console CSP (#2051)

`script-src` is nonce-only, fresh per request, stamped on the wasm bootstrap. Five tests including *the rendered nonce is the one the header authorizes* (they must agree or the page is broken, not merely unprotected) and *every response carries a fresh nonce*. `style-src` keeps its inline allowance and the test **says so by name** rather than leaving it to be discovered.

## Chart values that boot (#2159) — the release blocker

None of the committed `deploy/helm/ci/` files produced an acceptable configuration: no auth mechanism, a placeholder failing the Argon2 PHC check, a 15-byte HMAC against the RFC 8725 32-byte floor. These represent our own recommended deployments and they crash-loop; the tag lane boots them, so this failed the next cut. It survived because `validate.sh` renders and lints and nothing ever *started* one — `boot-check.sh` now does, on every PR rather than only on a tag.

## Gates

- `cargo clippy -p ferroehr -p ferroehr-rest --all-targets --all-features -- -D warnings` — clean
- `cargo clippy -p ferroehr-admin-ui` on **both** targets (ssr + wasm32) — clean
- `cargo nextest run -p ferroehr-admin-ui --features ssr` — **277 passed** (the feature matters: without it the nonce tests do not compile, which is also how CI runs it)
- `cargo nextest run -p ferroehr -E 'test(/audit_chain/)'` — 8 passed against a real database
- `cargo nextest run -p ferroehr -E 'test(/signing_pgp/)'` — 14 passed
- `comment-style`, `typed-status`, `default-style`, `error-hygiene`, `sql-string-building`, `docs-claims`, `image-labels` — all OK at `--all`
- `deploy/helm/validate.sh` — all render checks passed
- Full `--workspace` run in progress at time of writing; one earlier failure (`events_amqp::broker_down_then_up_delivers_without_loss`) was **resource contention** from ten live validation stacks, and passes in isolation — recorded rather than dismissed as flaky.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,7 @@ jobs:
       admin_ui: ${{ steps.filter.outputs.admin_ui }}
       server_dockerfile: ${{ steps.filter.outputs.server_dockerfile }}
       helm: ${{ steps.filter.outputs.helm }}
+      chart_boot: ${{ steps.filter.outputs.chart_boot }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
@@ -98,13 +99,25 @@ jobs:
               echo "admin_ui=true"
               echo "server_dockerfile=true"
               echo "helm=true"
+              echo "chart_boot=true"
             } >> "$GITHUB_OUTPUT"
             exit 0
           fi
           changed=$(git diff --name-only "$base" "$head")
           echo "Changed files:"; echo "$changed"
+          # Every output is written here for the reason the fallback above
+          # states: an unset output reads as `false` downstream and a skipped
+          # job counts as satisfied, so writing only `code` would leave the
+          # other gates unset rather than deliberately false.
           if [ -z "$changed" ]; then
-            echo "code=false" >> "$GITHUB_OUTPUT"; exit 0
+            {
+              echo "code=false"
+              echo "admin_ui=false"
+              echo "server_dockerfile=false"
+              echo "helm=false"
+              echo "chart_boot=false"
+            } >> "$GITHUB_OUTPUT"
+            exit 0
           fi
           # scripts/ is NOT build-relevant by default: the cargo jobs never
           # execute repo scripts, so a new watcher/site/release script must
@@ -147,6 +160,21 @@ jobs:
             echo "helm=true" >> "$GITHUB_OUTPUT"
           else
             echo "helm=false" >> "$GITHUB_OUTPUT"
+          fi
+          # Chart BOOT gate (#2159), a strict superset of the helm filter: the
+          # overlays are only half of the pair under test, the other half being
+          # the configuration vocabulary they are written against. Both sides
+          # can break it — the overlays went un-bootable because the SERVER
+          # tightened validation (an auth mechanism became mandatory, the HMAC
+          # floor arrived, a password hash started being parsed at boot) while
+          # the values files stood still and every render stayed green. So a
+          # change to the config tree, its validators or the annotated template
+          # that is the schema of record re-boots the overlays too. The
+          # Dockerfile is in because it decides what the binary is packaged into.
+          if echo "$changed" | grep -qE '^(deploy/helm/|\.github/workflows/ci\.yml$|docker/Dockerfile$|app/ferroehr/src/config/|app/ferroehr/assets/ferroehr\.default\.toml$|app/ferroehr-rest/src/config\.rs$)' ; then
+            echo "chart_boot=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "chart_boot=false" >> "$GITHUB_OUTPUT"
           fi
 
   # ── Required for code changes: formatting ───────────────────────────────────
@@ -741,6 +769,70 @@ jobs:
       - name: Chart validation + golden renders
         run: bash deploy/helm/validate.sh
 
+  # ── Every committed values overlay must BOOT, not merely render (#2159) ─────
+  # helm-golden above proves the manifests are well-formed, hardened and
+  # unchanged. It proved exactly that while all three overlays carried
+  # configurations the server refuses to start on: `auth.enabled` with no
+  # mechanism, a placeholder that is not an Argon2id PHC string, a 15-byte HMAC
+  # secret under a 32-byte floor, SMART enabled with no public base URL. Four
+  # boot failures, four green renders. Nothing but the server can tell the
+  # difference, so this job runs it.
+  #
+  # It runs on every PR that can break the pair (see the `chart_boot` filter),
+  # NOT only on a tag — the tag lane's identical check is the last line of
+  # defence, and by then the release cut is already failing.
+  #
+  # The image is packaged from THIS TREE's server binary rather than pulled from
+  # GHCR, which is the difference between a gate and a trap: `appVersion` and the
+  # chart's config defaults advance on different clocks, so judging a PR against
+  # a published image would fail every PR that adds a configuration key, from the
+  # commit that adds it until the next release. The binary comes from the
+  # ui-e2e-binary job, whose exact-key cache makes a chart-only PR pay nothing
+  # for it, and packaging is the COPY-only `runtime-prebuilt` target — seconds.
+  chart-boot:
+    name: chart-boot
+    needs: [changes, ui-e2e-binary]
+    if: needs.changes.outputs.chart_boot == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      # The same pin as the goldens, from the same file: `azure/setup-helm` takes
+      # `version`, never `version-file`, and an unknown input is only a warning.
+      - name: Read the pinned helm version
+        id: helm-pin
+        run: |
+          set -euo pipefail
+          version="$(awk '$1 == "helm" { print $2 }' deploy/helm/.tool-versions)"
+          [ -n "$version" ] || { echo "::error::no helm line in deploy/helm/.tool-versions"; exit 1; }
+          echo "version=v${version}" >> "$GITHUB_OUTPUT"
+      - uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5.0.1
+        with:
+          version: ${{ steps.helm-pin.outputs.version }}
+      # The artifact download loses the executable bit; the Dockerfile COPYs the
+      # binary with `--chmod=0755`, which is why that does not matter here.
+      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: ui-e2e-server-binary
+          path: bin/amd64
+      - uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+      - name: Package the app image (COPY-only, seconds)
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
+        with:
+          context: .
+          file: docker/Dockerfile
+          target: runtime-prebuilt
+          load: true
+          tags: ferroehr:chart-boot
+      - name: Boot every committed values overlay
+        env:
+          FERROEHR_IMAGE: ferroehr:chart-boot
+        run: bash deploy/helm/ci/boot-check.sh
+
   # ── Unused dependencies ──────────────────────────────────────────────────────
   unused-deps:
     name: cargo-machete
@@ -794,10 +886,15 @@ jobs:
   # build.rs degrades to SHA=unknown BY DESIGN, and the battery asserts
   # behaviour, never provenance. The PUBLISHED images are untouched:
   # containers.yml still stamps github.sha into them.
+  #
+  # Two consumers now (#2159): the ui-e2e battery below, and the `chart-boot`
+  # job, which boots the chart's values overlays against an image packaged from
+  # this same binary. Hence the widened condition — the job id and name are
+  # unchanged on purpose, because they are branch-protection check names.
   ui-e2e-binary:
     name: ui-e2e server binary
     needs: changes
-    if: needs.changes.outputs.admin_ui == 'true'
+    if: needs.changes.outputs.admin_ui == 'true' || needs.changes.outputs.chart_boot == 'true'
     runs-on: ubuntu-latest
     # Generous: a binary-cache miss with a cold sccache is a full release
     # build of the ~190 kLoC first-party graph (714 s measured cold, plus
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,62 @@ workflow refuses a tag that has no matching section here.
 
 ### Added
 
+- **The PGP signing key can be rotated without losing history.** Signing now
+  uses a signing-capable *subkey* selected by its OpenPGP key flags, which is
+  the rotation mechanism the standard provides: issue a new subkey, the
+  certificate retains the old one, and every previously-signed version keeps
+  verifying with no configuration change. For a certificate that is genuinely
+  replaced — a compromised key, an organisational change — `signing.retired_key_paths`
+  holds the retired **public** keys, verify-only, so a retired key can never
+  sign again. Verification also no longer ignores subkeys, which it did before.
+
+- **`deploy/helm/ci/boot-check.sh`** — runs a real ferroehr image against every
+  values overlay the chart ships, replaying the complete delivery surface: the
+  rendered `ferroehr.toml`, every `config.files` entry, every file-borne secret
+  at its real value, and the environment the Deployment declares (including
+  `valueFrom.secretKeyRef`). Point `FERROEHR_IMAGE` at the tag you intend to
+  deploy and pass your own values file to check it before installing. (#2159)
+- **CI lane `chart-boot`** — boots every committed values overlay against an
+  image packaged from the branch under test, on every pull request that touches
+  the chart or the configuration vocabulary the overlays are written against.
+  Previously the equivalent check ran only on a release tag, which is after the
+  cut has already failed. (#2159)
+
+- **The server can run with no DDL rights at all.** `db.migrate = "verify"`
+  makes it issue no schema statements: at boot it checks that the database
+  already carries exactly this build's migrations and refuses to start
+  otherwise, naming the schema and what is wrong with it (never migrated,
+  behind this build, ahead of it, a failed migration, or a migration applied
+  from different source text). The runtime DSN can then authenticate as
+  `ferroehr_app` alone, so an application-level SQL flaw reaches rows and never
+  the schema. `apply` remains the default, so an empty configuration still
+  boots against an empty database. (#2049)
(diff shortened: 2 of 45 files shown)


rubentalstra:
Full workspace run is green with this batch in: `cargo nextest run --workspace --no-fail-fast` → **4447 passed, 0 failed, 6 skipped** (443s).

That settles the one open question in the PR body: the earlier `events_amqp::broker_down_then_up_delivers_without_loss` failure was resource contention from ten live validation stacks running alongside, not a defect. It passed in isolation and it passes in the full run now that the machine is quieter. Recorded rather than written off as flaky, because "flaky" is where real failures go to hide.

Issue #2164, opened 2026-08-07
test(deploy): the admin console against a live CDR, including its OIDC login


rubentalstra:
The console has E2E journeys (`scripts/ui-e2e.sh`) against a composed stack. What is not demonstrated is the **deployment**: its own OCI image, its own Helm/compose wiring, and its OIDC login against a real identity provider.

What must be shown:

- The console image deployed alongside the CDR — in compose (the `admin-ui` profile) and on Kubernetes — reaching the CDR **only over ITS-REST**, as its mandate requires.
- Logging in through the live Keycloak from #2158's Keycloak work, not a fixture session.
- The main journeys working against real data: template upload, an AQL query through the builder, browsing an EHR, viewing a composition.
- Its CSP behaving on the deployed image (see the nonce work in #2051) — the header observed on a real response.
- The console turned OFF leaving a working CDR with no residue.

## Acceptance criteria
- [x] The console deployed and driven on both compose and Kubernetes, transcript recorded
- [x] A real OIDC login against the live identity provider
- [x] The main journeys exercised against real data, with any failure filed
- [x] Zero browser-console errors during the run — the journeys already fail on hydration errors, and that property must hold on the deployed image too

rubentalstra:
Deprioritized — the server comes first (standing rule).

Probes for the console's DEPLOYMENT are written and parked on `test/console-deployment-probes` rather than merged: the shipped CSP on a real response (nonce-based `script-src`, no `unsafe-inline` beside it, a fresh nonce per request), the REST-only boundary observed as the container carrying no DSN, and the off-state leaving the CDR untouched. They are unverified, because verifying them needs a console image built from develop — the published 3.17.3 image predates the #2051 nonce work — and that build is not where this milestone's effort belongs.

The browser journeys remain `scripts/ui-e2e.sh`'s job and already run against the composed console image. This issue stays open for the Kubernetes half and the live OIDC login.

rubentalstra:
## Deployment transcript — the console driven on compose and Kubernetes

### Compose lane (the shipped-artifact battery, OIDC included)

`UI_E2E_IMAGE=1 bash scripts/ui-e2e.sh` — the console COMPOSED as its own OCI image (built by `docker/admin-ui/Dockerfile` through the compose override) beside the CDR image, Keycloak (the full exported realm) and postgres, on fresh volumes:

[...]

PR #2195, opened 2026-08-07
fix(chart,observability)!: probes that mean what they say, the console as a real workload, and metrics that arrive

### LINKED ISSUE #2066: security(authz): per-object EHR access defaults to permit, with no server-wide way to change it
Found by checking the surface against the OWASP **Insecure Direct Object Reference Prevention** Cheat Sheet: *"implement access control checks for each object"*, *"verify the user's permission every time an access attempt is made"*, and the caveat that matters most here — *"even with complex identifiers, access control checks are essential."*

`ehr_id` is a UUIDv7, which is exactly the unpredictable identifier the sheet says is **not** sufficient on its own.

The gate exists and runs first on every `/ehr/{ehr_id}` route. Its default disposition is permit: an EHR with no `ACCESS_CONTROL_SETTINGS` returns `Ok(())`, and `DefaultAccess::Open` does the same. A newly created EHR carries no settings, so **any authenticated caller holding the clinical role reaches any EHR by id.** The finer layer that would bind caller to patient (ABAC) is off by default.

Two things stated fairly, because they make this a posture rather than a bug: it is **documented honestly** — the security page says "by default it has no settings and the EHR is open to any authenticated caller" — and openEHR itself places authorization out of band (SM `master02` §General Assumptions), which the code cites.

**The actionable gap is narrower than "IDOR".** `default_access: restricted` is settable only **per EHR, through a CONTRIBUTION commit per record**. A deployment that wants object-level default-deny has no configuration to express it — it would have to author a settings object for every EHR it ever creates, including ones created by clients. That asymmetry is the defect: the safe posture is reachable only by the most laborious path.

## Acceptance criteria
- [ ] A server-wide default applied when an EHR carries no `ACCESS_CONTROL_SETTINGS` (a `[authz]` key), so a deployment can choose default-deny once
- [ ] The default for new deployments decided explicitly and justified — changing it is a behaviour change for existing ones, so the migration story is part of this
- [ ] Tests for both dispositions, including that an EHR with explicit settings still overrides the server default
- [ ] The book stating which posture a deployment is running and how to change it

### LINKED ISSUE #2165: docs(deploy): enabling an optional integration must be a small, findable change
Correctness property, not polish: an integration an operator cannot discover and switch on from the documentation is not really shipped.

For **each** of SeaweedFS/multimedia, Keycloak/OIDC, observability, signing (digest and PGP), the admin console, FHIR, events, tenancy and terminology, establish by following the book alone — not the source:

- Where the reader is told the integration exists.
- The smallest change that enables it, in **both** compose and Helm.
- What else they must supply (a secret, a URL, a key), stated up front rather than discovered from a boot failure.
- How to turn it off again.

Anything that takes more than a small documented change, or that requires knowing a key the book never mentions, is a finding: fix the ergonomics or fix the documentation, and say which.

Note one shape already suspected: the chart renders `config:` verbatim to TOML, so every server key is reachable as `config.<toml.path>` whether or not `values.yaml` enumerates it. That is powerful and nearly invisible — the book should say it plainly, because it is the difference between "unsupported" and "already supported, undocumented".

## Acceptance criteria
- [ ] Each integration has a documented enable/disable path for compose and for Helm, verified by following it
- [ ] Every prerequisite stated before the reader hits a boot failure
- [ ] The `config.*` passthrough documented as the general mechanism it is
- [ ] Any integration that cannot be enabled simply is fixed or its difficulty is stated honestly

### LINKED ISSUE #2168: fix(multimedia,compose): the documented SeaweedFS setup 500s on the first commit — the bucket is never created
Following the published documentation for S3 multimedia end to end produces a `500 Internal Server Error` on the very first composition that carries a large `DV_MULTIMEDIA`. **The bucket is never created** — neither by the `seaweedfs` Compose service, nor by the server, nor by any step the book asks the operator to run — and SeaweedFS answers a `PUT` into a missing bucket with `403 AccessDenied`, which sends the operator hunting for credentials that are not the problem.

The book says the gateway "needs no credentials" (`website/book/src/beyond-core/s3-multimedia.md` §Quick setup with SeaweedFS) and the Compose file says it "runs in unauthenticated allow-all mode". Both are true. The missing piece is the bucket.

## Reproduction

`ghcr.io/rubentalstra/ferroehr:3.17.3` (`git_sha 5a9557bf3a3d7fb12004eff94db74c342634f8d5`), Docker 29.6.2 / Compose v5.3.1, quickstart `docker-compose.yml` + `--profile s3`, `chrislusf/seaweedfs:4.40`, multimedia configured exactly as the book documents (`ENABLED=true`, `ENDPOINT=http://seaweedfs:8333`, `BUCKET=openehr-multimedia`, `ALLOW_HTTP=true`), confirmed live:

```
$ curl -s -u ferroehr:ferroehr http://localhost:18080/management/env | jq .multimedia
{"enabled": true, "threshold_bytes": 262144, "endpoint": "http://seaweedfs:8333",
 "bucket": "openehr-multimedia", "region": "us-east-1", "access_key_id": null,
 "secret_access_key": "***", "secret_access_key_file": "***", "allow_http": true}
```

Commit a composition with 400 KiB of inline `DV_MULTIMEDIA.data`:

```
$ curl -s -o /tmp/mm_resp.json -w 'status=%{http_code}\n' -u ferroehr:ferroehr -X POST \
    -H 'Content-Type: application/json' --data-binary @comp_mm.json \
    http://localhost:18080/ferroehr/rest/openehr/v1/ehr/$EHR/composition
status=500
{"error":"Internal Server Error","message":"internal error: the server encountered an internal error"}
```

Server log:

```
ERROR ferroehr::service::error: internal server fault → 500 context="complete the request"
  error=blob store: The operation lacked the necessary privileges to complete for path
  fac4ffbf…557d: Error performing PUT http://seaweedfs:8333/openehr-multimedia/fac4ffbf…557d
  in 6.205375ms - Server returned non-2xx status code: 403 Forbidden:
  <Error><Code>AccessDenied</Code><Message>Access Denied.</Message>…</Error>
```

The gateway is reachable and anonymous access works — the bucket simply does not exist:

```
$ curl -s http://localhost:18333/
<ListAllMyBucketsResult …><Owner><ID></ID></Owner><Buckets></Buckets></ListAllMyBucketsResult>   # empty

$ curl -s -i -X PUT --data hello http://localhost:18333/openehr-multimedia/probe1 | head -1
HTTP/1.1 403 Forbidden                                    # missing bucket → AccessDenied, not NoSuchBucket

$ curl -s -i -X PUT http://localhost:18333/openehr-multimedia | head -1
HTTP/1.1 200 OK                                           # anonymous bucket creation is allowed

$ curl -s -i -X PUT --data hello http://localhost:18333/openehr-multimedia/probe1 | head -1
HTTP/1.1 200 OK                                           # anonymous writes work once the bucket exists
```

With the bucket created by hand, the identical commit returns `201` and the blob lands under its SHA-256 key. So the only missing step is bucket creation, and nothing in the shipped artifacts or the book performs it.

## Two separable defects

1. **Setup gap.** Nothing creates the bucket. Options: a `seaweedfs` bucket-init in the Compose file, a documented one-liner, or the server creating the bucket on demand when it is absent.
2. **Diagnosis gap.** An operator sees `internal error: the server encountered an internal error` on the wire and `Access Denied` in the log for what is a missing-bucket condition. A permanent misconfiguration is reported as a transient-looking `500` on every subsequent write.

## Interim documentation fix already landed
`website/book/src/beyond-core/s3-multimedia.md` and `website/book/src/installation/compose.md` gain the explicit bucket-creation step, so an operator following the book now gets a working system. That does not close this issue — the shipped stack should not need a manual step.

## Acceptance criteria
- [ ] Following the documented SeaweedFS setup with no extra manual step yields a successful multimedia commit
- [ ] A missing/unreachable bucket produces an operator-actionable failure rather than a bare `internal error`
- [ ] Evidence recorded from a live stack, not from a render

### LINKED ISSUE #2169: fix(compose,docs): FERROEHR__MULTIMEDIA__* host env is silently ignored — the documented s3 recipe does nothing
`website/book/src/installation/compose.md` §Optional services tells the operator:

> Start it with `docker compose --profile s3 up`, then set `FERROEHR__MULTIMEDIA__ENABLED=true`, `FERROEHR__MULTIMEDIA__ENDPOINT=http://seaweedfs:8333`, …

and the header comment of `docker-compose.yml` says the same. The natural reading — export them in the shell — **has no effect at all**: the `ferroehr` service's `environment:` block does not pass those keys through, and Compose forwards nothing implicitly. The stack comes up healthy and multimedia stays off, silently.

Every other tunable the file documents (`FERROEHR__LOG__FORMAT`, `FERROEHR__DB__MAX_CONNECTIONS`, `FERROEHR__SIGNING__ENABLED`, …) *is* wired as a `${VAR:-default}` pass-through, so the multimedia keys are the odd ones out.

## Reproduction

`ghcr.io/rubentalstra/ferroehr:3.17.3`, Docker 29.6.2 / Compose v5.3.1, the standalone quickstart `docker-compose.yml` in an empty directory (no repo override merged).

```
$ export FERROEHR__MULTIMEDIA__ENABLED=true
$ export FERROEHR__MULTIMEDIA__ENDPOINT=http://seaweedfs:8333
$ export FERROEHR__MULTIMEDIA__BUCKET=openehr-multimedia
$ export FERROEHR__MULTIMEDIA__ALLOW_HTTP=true
$ env | grep MULTIMEDIA | sort
FERROEHR__MULTIMEDIA__ALLOW_HTTP=true
FERROEHR__MULTIMEDIA__BUCKET=openehr-multimedia
FERROEHR__MULTIMEDIA__ENABLED=true
FERROEHR__MULTIMEDIA__ENDPOINT=http://seaweedfs:8333

$ docker compose --profile s3 up -d --wait ferroehr seaweedfs
… Container live2161-ferroehr-1  Healthy

$ docker compose exec ferroehr env | grep -i multimedia
(none)

$ curl -s -u ferroehr:ferroehr http://localhost:18080/management/env | jq -c .multimedia
{"enabled":false,"threshold_bytes":262144,"endpoint":null,"bucket":"openehr-multimedia",
 "region":"us-east-1","access_key_id":null,"secret_access_key":"***",
 "secret_access_key_file":"***","allow_http":false}
```

The operator has done exactly what the book says and the integration is off, with no warning anywhere.

## The fix is not purely mechanical

A `${FERROEHR__MULTIMEDIA__ENDPOINT:-}` pass-through in the map syntax the file uses sets the variable to the **empty string** when the operator does not set it, and empty values are not benign:

```
$ docker run --rm -e FERROEHR__DB__URL=… -e FERROEHR__MULTIMEDIA__ALLOW_HTTP= ghcr.io/rubentalstra/ferroehr:3.17.3
Error: 1 configuration error(s):
  - invalid configuration: invalid type: string "", expected a boolean for key `multimedia.allow_http` in the environment
```

and an empty `endpoint` panics the request handler (filed separately as the multimedia-endpoint panic issue). So the pass-through needs real defaults per key (`false` / `openehr-multimedia` / `us-east-1` / `262144`), and `endpoint` needs the panic fixed or a non-empty default before it can be defaulted safely.

## Interim documentation fix already landed
`website/book/src/installation/compose.md` now shows the working recipe (adding the keys to the `ferroehr` service's `environment:` block) instead of the misleading "then set …". That removes the trap; it does not give back the one-command ergonomics the parent issue asks for.

## Acceptance criteria
- [ ] `FERROEHR__MULTIMEDIA__ENABLED=true docker compose --profile s3 up` (or whatever the book ends up documenting) actually enables the integration, verified on a live stack via `/management/env`
- [ ] No key defaults to a value that fails config parsing or panics a handler
- [ ] The book and the Compose file header describe the same, working, mechanism

### LINKED ISSUE #2170: ci(publish-chart): the chart acceptance gate replays placeholder secrets and judges only one overlay
The tag lane's acceptance step (`.github/workflows/publish-chart.yml`, step *"The selected image must accept the chart's rendered defaults"*) is the last gate before a chart is published, and it is weaker than it reads. Three gaps, all of which let a non-bootable chart through:

**1. It writes a placeholder into every file-borne secret, so secret CONTENT is never judged.** The replay loop writes `printf 'x'` into every `*_FILE` path (a DSN for `DB__URL_FILE`, `x` for everything else). But several boot rules are rules about the value: `auth.oidc.hmac_secret` has a 32-byte floor (RFC 8725 §3.5), `auth.basic.users[].password_hash` must parse as an Argon2id PHC string, `signing.key_passphrase` is read. #2159's 15-byte `dev-hmac-secret` is invisible to this step by construction — `x` is short too, but the step never renders the overlay that carries it.

**2. It judges only `default-values.yaml`.** `all-features-values.yaml` and `basic-auth-values.yaml` are published in the same chart and were both un-bootable (#2159). The overlay with every integration enabled is the one most likely to break.

**3. It replays only literal `value:` environment.** The awk pass matches `- name: FERROEHR__…` followed by `value:`; a `valueFrom.secretKeyRef` entry is skipped. The all-features render has two (`FERROEHR__AUDIT__FHIR_FEED__URL`, `FERROEHR__MULTIMEDIA__ACCESS_KEY_ID`), so those keys are never presented to the strict env sweep the step exists to exercise.

## Reproduction

The gap in (1) is visible by running the step's own logic against the pre-fix all-features overlay: with `printf 'x'` in the HMAC file the server reports the length of the PLACEHOLDER, not of the configured secret.

```
$ # publish-chart.yml's replay: 'x' into every *_FILE path
$ docker run --rm -v .../rendered.toml:/etc/ferroehr/ferroehr.toml:ro \
    -v .../envfiles:/etc/ferroehr-secrets:ro [env...] \
    --entrypoint /usr/local/bin/ferroehr ghcr.io/rubentalstra/ferroehr:develop config check
Error: 2 configuration error(s):
  - auth: auth.oidc.hmac_secret is 1 bytes; a keyed-MAC key must be at least 32 bytes ...
  - smart: smart.public_base_url ... is required when SMART is enabled
```

`1 bytes` is the `x`. The overlay's real value was 15 bytes. Both are refusals here only by luck of both being short; a placeholder longer than 32 bytes would have reported success against a 15-byte configured secret.

## Resolution

`deploy/helm/ci/boot-check.sh` (added in #2159) already does all three correctly: it splits the rendered Secret and mounts each key at its REAL value, resolves `valueFrom.secretKeyRef`, and enumerates every committed overlay rather than naming one. Replace the hand-rolled replay in `publish-chart.yml` with a call to it:

```yaml
- name: The selected image must accept every committed values overlay
  env:
    FERROEHR_IMAGE: ${{ steps.judge.outputs.image }}
  run: bash deploy/helm/ci/boot-check.sh
```

That also removes the second copy of the replay logic, which is how the two drifted apart in the first place.

## Acceptance criteria
- [ ] The tag lane's acceptance step boots every committed overlay, not only `default-values.yaml`
- [ ] Secret VALUES are the rendered ones, so a too-short HMAC or a non-PHC password hash fails the gate
- [ ] `valueFrom.secretKeyRef` environment is replayed
- [ ] The replay logic exists in exactly one place

### LINKED ISSUE #2171: fix(multimedia): disabling the integration silently ignores ?expand_multimedia=true and strands offloaded blobs
Turning `[multimedia] enabled` back off strands every blob that was already externalized: `?expand_multimedia=true` is **silently ignored**. The read answers `200` with the compact `s3://` reference and no indication that the expansion the client asked for did not happen. The bytes are still in the object store and still reachable by an operator with an S3 client, but the API will not return them and will not say so.

This is the "turn each integration off cleanly, no residue" property from the parent programme, and it is the one place where turning something off is not clean.

## Reproduction

`ghcr.io/rubentalstra/ferroehr:3.17.3`, quickstart Compose + `--profile s3`, `chrislusf/seaweedfs:4.40`.

With `FERROEHR__MULTIMEDIA__ENABLED=true`, commit a composition carrying 400 KiB of inline `DV_MULTIMEDIA.data`. It offloads, and the expanded read returns the bytes:

```
$ curl -s -u ferroehr:ferroehr "$B/ehr/$EHR/composition/$CUID?expand_multimedia=true" -w '%{http_code} %{size_download}\n' -o r.json
200 1116966
   → decoded 409600 bytes, sha256 fac4ffbf…557d, byte-identical to what was committed
```

Now set `FERROEHR__MULTIMEDIA__ENABLED=false` and restart the server (nothing else changed — the endpoint, bucket and store are all still configured and running):

```
$ curl -s -u ferroehr:ferroehr http://localhost:18080/management/env | jq -c .multimedia
{"enabled":false,"threshold_bytes":262144,"endpoint":"http://seaweedfs:8333","bucket":"openehr-multimedia",…}

$ curl -s -u ferroehr:ferroehr "$B/ehr/$EHR/composition/$CUID?expand_multimedia=true" -w '%{http_code}\n' -o r2.json
200
   → size=409600, inline data = False, uri = s3://openehr-multimedia/fac4ffbf…557d
```

The blob is still in the store:

```
$ curl -s "http://localhost:18333/openehr-multimedia?list-type=2" | grep -o '<Key>[^<]*</Key>'
<Key>85ff460b12e808be50432bf9f913b2cf742e1c917aa09d9763faf28e45c36864</Key>
<Key>fac4ffbf2c780761c9ffbaa7529982cb99181c59b4be27487af6e4de1192557d</Key>
```

The commit half of "disabled" is fine and worth recording: with the integration off, a 400 KiB `DV_MULTIMEDIA` is stored inline, `201`, byte-identical, with no dependency on the store at all (`ehr.node` grows to hold it — 4 nodes above 300 KB, 1212 kB total after the test).

## Three honest options, one of which must be chosen
1. Keep the fetch/expand path alive whenever an endpoint is configured, so `enabled` governs *new offloads* only.
2. Refuse `expand_multimedia=true` with a typed error when the fetch path is off, so the client learns the request was not honoured.
3. Keep the current silence — but then the book must state it plainly, and it is still a silent data-access loss.

Silently ignoring a query parameter that changes what clinical content the caller receives is the failure class this project's reliability rules exist to prevent.

## Acceptance criteria
- [ ] The behaviour of `expand_multimedia=true` with the integration disabled is deliberate rather than incidental, and the client can tell what happened
- [ ] The documented lifecycle covers what happens to already-externalized blobs when the integration is switched off
- [ ] Observed live, not inferred

### LINKED ISSUE #2173: fix(compose,observability): the bundled Prometheus scrapes nothing — no FerroEHR metric ever reaches the observability stack
`docker-compose.observability.yml` says, in its own header, that metrics are "pulled by the bundled Prometheus from ferroehr:9464/management/prometheus", and `website/book/src/installation/compose.md` repeats it. **No metric ever arrives.** The bundled Prometheus has zero scrape jobs, so `ferroehr_build_info` and every other server metric are absent, and the provisioned Grafana dashboard's metric panels have nothing behind them.

Traces are fine — this is only the metrics half.

## Reproduction

`ghcr.io/rubentalstra/ferroehr:develop` (`git_sha 1785c2ef9ac32f3636d81936829acd4304c3bc39`, built 2026-08-07T09:43:33Z), `grafana/otel-lgtm:0.29.2` (digest-pinned in the overlay), Docker 29.6.2 / Compose v5.3.1.

```
$ docker compose -f docker-compose.yml -f docker-compose.observability.yml up -d --wait ferroehr otel-lgtm
 Container live2162-otel-lgtm-1  Healthy
 Container live2162-ferroehr-1   Healthy
```

The server side is entirely correct — the endpoint serves the metric with the right identity:

```
$ docker compose … exec -T otel-lgtm sh -lc 'curl -s http://ferroehr:9464/management/prometheus | grep "^ferroehr_build_info"'
ferroehr_build_info{version="3.17.3",git_sha="1785c2ef9ac32f3636d81936829acd4304c3bc39",rm_version="1.2.0"} 1

$ … curl -s http://ferroehr:9464/management/info
{"name":"ferroehr","version":"3.17.3","git_sha":"1785c2ef9ac32f3636d81936829acd4304c3bc39",…}
```

The overlay's inline scrape config is mounted exactly where the overlay puts it:

```
$ … exec -T otel-lgtm sh -lc 'ls -la /otel-lgtm/prometheus/scrape-configs/; cat /otel-lgtm/prometheus/scrape-configs/ferroehr.yaml'
-r--r--r-- 1 root root 180 Aug  7 11:15 ferroehr.yaml
- job_name: ferroehr
  metrics_path: /management/prometheus
  scrape_interval: 15s
  static_configs:
    - targets:
        - ferroehr:9464
      labels:
        service: ferroehr
```

And Prometheus never reads it:

```
$ … curl -s http://localhost:9090/api/v1/targets
{"status":"success","data":{"activeTargets":[],"droppedTargets":[],"droppedTargetCounts":{}}}

$ … curl -s 'http://localhost:9090/api/v1/query?query=ferroehr_build_info'
{"status":"success","data":{"resultType":"vector","result":[]}}

$ … curl -s http://localhost:9090/api/v1/label/__name__/values
{"status":"success","data":["otelcol_exporter_in_flight_requests","otelcol_exporter_queue_capacity",…]}
   → only otelcol_* series exist; nothing from FerroEHR
```

## Root cause, read out of the image

`grafana/otel-lgtm:0.29.2` starts Prometheus with one fixed config file and no include of that directory:

```
$ … cat /otel-lgtm/run-prometheus.sh
run_with_logging "Prometheus ${PROMETHEUS_VERSION}" … ./prometheus/prometheus \
	--web.enable-remote-write-receiver \
	--web.enable-otlp-receiver \
	--enable-feature=exemplar-storage \
	--storage.tsdb.path=/data/prometheus \
	--config.file=./prometheus.yaml \
```

`./prometheus.yaml` is `/otel-lgtm/prometheus.yaml`, and it contains `otlp:`, `storage:` and `global:` — **no `scrape_configs`, no `scrape_config_files`**. (`/otel-lgtm/prometheus/prometheus.yml`, which does have a `scrape_configs` block, is a leftover sample the process never loads — confirmed by comparing `/api/v1/status/config` against both files.) In this image metrics reach Prometheus over OTLP from the collector, whose only scrape job is its own `127.0.0.1:8888`:

```
$ … grep -A6 'prometheus/collector' /otel-lgtm/otelcol-config.yaml
  prometheus/collector:
    config:
      scrape_configs:
        - job_name: "opentelemetry-collector"
          scrape_interval: 1s
          static_configs:
            - targets: ["127.0.0.1:8888"]
```

So the overlay mounts a file into a directory nothing reads.

## Two candidate fixes, and the server already has one of them

The server exposes a **push** path that the collector's OTLP receiver would accept directly, and the overlay leaves it off:

```
$ … curl -s -u ferroehr:ferroehr http://ferroehr:9464/management/env | jq .telemetry
{"otlp_endpoint":"http://otel-lgtm:4317","service_name":"ferroehr","environment":"dev",
 "traces_sample_ratio":1.0,"metrics_push":false,"flame_file":null}
```

1. Set `FERROEHR__TELEMETRY__METRICS_PUSH=true` and let metrics ride the OTLP connection that already carries traces successfully. This matches how the image is designed to receive metrics and drops the scrape config entirely. **Needs verifying live** — this issue does not claim it works, only that it exists.
2. Keep the pull model and mount a full replacement `/otel-lgtm/prometheus.yaml` carrying the image's own `otlp:`/`storage:` blocks plus a `scrape_configs` entry. Works, but couples the overlay to the image's internal config file.

Whichever is chosen, the header comment in the overlay and the paragraph in `website/book/src/installation/compose.md` must describe what actually happens.

## Acceptance criteria
- [ ] `ferroehr_build_info` is queryable in the bundled Prometheus with the running build's version and SHA, observed live
- [ ] The provisioned Grafana dashboard's metric panels render data rather than "No data"
- [ ] The overlay's own comment and the book describe the mechanism that is actually in use

### LINKED ISSUE #2176: fix(compose,keycloak): the documented OIDC quickstart returns 401 — the demo realm never puts `ferroehr` in `aud`
The OIDC quickstart in `website/book/src/installation/compose.md` §The OIDC variant hands the reader two commands: fetch a token from the bundled Keycloak, then call the API with it. **The second one returns `401`.** The demo realm never puts `ferroehr` in the token's `aud`, and the same overlay configures the server to require exactly that audience — so the two halves of one shipped file contradict each other, and the documented flow cannot succeed as written.

## Reproduction

`ghcr.io/rubentalstra/ferroehr:develop` (`git_sha 1785c2ef9ac32f3636d81936829acd4304c3bc39`), `quay.io/keycloak/keycloak:26.7.0` (the digest pinned in the overlay), Docker 29.6.2 / Compose v5.3.1, the standalone `docker-compose.yml` + `docker-compose.keycloak.yml` in an empty directory.

The overlay requires the audience:

```yaml
FERROEHR__AUTH__OIDC__AUDIENCES: ferroehr
```

The realm it ships declares one client and one user, the user carrying an explicit `realmRoles` list. Both token shapes that definition can produce were exercised, and both are refused:

**(a) user with an explicit `realmRoles` list — what the realm import declares.** No `aud` claim is emitted at all:

```
$ curl -s -d client_id=ferroehr -d client_secret=ferroehr-quickstart-secret \
    -d username=clinician -d password=pw -d grant_type=password \
    http://localhost:18081/auth/realms/ferroehr/protocol/openid-connect/token | jq -r .access_token
   → decoded claims:
     "iss": "http://keycloak:8081/auth/realms/ferroehr",
     "azp": "ferroehr",
     "realm_access": {"roles": ["USER"]},
     "scope": "profile email"
     (no "aud" key)

$ curl -s -i -X POST -H "Authorization: Bearer $T" http://localhost:18080/ferroehr/rest/openehr/v1/ehr
HTTP/1.1 401 Unauthorized
www-authenticate: Basic realm="ferroehr", charset="UTF-8", Bearer realm="ferroehr", error="invalid_token"
{"error":"Unauthorized","message":"unauthorized: invalid bearer token: Missing required claim: aud"}
```

**(b) user that also carries the realm's default roles.** `aud` is `account`, never `ferroehr`:

```
  aud = account   azp = ferroehr
  realm_access.roles = ['offline_access','uma_authorization','ADMIN','USER','default-roles-ferroehr']

$ curl -s -i -H "Authorization: Bearer $TOKEN" -X POST http://localhost:18080/ferroehr/rest/openehr/v1/ehr
HTTP/1.1 401 Unauthorized
{"error":"Unauthorized","message":"unauthorized: invalid bearer token: InvalidAudience"}
```

Keycloak only adds a client to `aud` when a protocol mapper says so, or when the token carries roles scoped to that client. The demo realm has neither, so no user of it can ever produce a token this server accepts.

The server is behaving correctly in both cases — RFC 7519 §4.1.3 audience validation is exactly right, and the refusal message is precise. **The realm is what is wrong.**

## The fix, proven live

Adding an `oidc-audience-mapper` to the `ferroehr` client makes the documented flow work, with nothing else changed:

```
$ curl -X POST …/clients/$CID/protocol-mappers/models -d '{
    "name":"ferroehr-audience","protocol":"openid-connect",
    "protocolMapper":"oidc-audience-mapper",
    "config":{"included.client.audience":"ferroehr","access.token.claim":"true","id.token.claim":"false"}}'
201

  aud = ['ferroehr', 'account']

$ curl -s -i -H "Authorization: Bearer $TOKEN" -X POST http://localhost:18080/ferroehr/rest/openehr/v1/ehr
HTTP/1.1 201 Created
location: /ferroehr/rest/openehr/v1/ehr/019fdc22-4e73-71c3-84ae-96df572e5597
```

The equivalent in the inline realm JSON is a `protocolMappers` entry on the client — landed on the branch for this sweep.

## Related, same file: the healthcheck budget is too small for a cold start

Not the same defect, recorded here because it is the same artifact and the same first-run experience. `docker compose -f docker-compose.yml -f docker-compose.keycloak.yml up` **fails outright on a cold start**: the Keycloak healthcheck allows `start_period 20s` + 30 × 10s ≈ 5m20s, the container took **9m31s** to answer (container started 10:55:08, health flipped at 11:04:39), and since `ferroehr` declares `depends_on: {keycloak: {condition: service_healthy}}` the CDR never starts at all:

```
 Container live2160-keycloak-1  Error dependency keycloak failed to start
dependency failed to start: container live2160-keycloak-1 is unhealthy

healthcheck log: /bin/sh: line 1: /dev/tcp/127.0.0.1/8081: Connection refused   (× 30)
keycloak log:    Quarkus augmentation completed in 144656ms
                 Initializing database schema. Using changelog META-INF/jpa-changelog-master.xml
```

Honest caveat: that timing was measured on a busy machine, so 9m31s is not a universal figure. But the budget only has to be short **once** for the documented command to fail, and the failure mode is the whole stack not starting.

## Acceptance criteria
- [ ] The two commands in the book's OIDC section work verbatim against the shipped overlay, verified live
- [ ] The demo realm emits `ferroehr` in `aud` for every user it defines
- [ ] A cold `up` of the overlay reaches a serving CDR without a second invocation

### LINKED ISSUE #2177: fix(docs,authz): authz.rbac.management_access is inert when per-endpoint levels are set, but the security chapter presents it as the gate
`website/book/src/security.md` §RBAC lists

| `FERROEHR__AUTHZ__RBAC__MANAGEMENT_ACCESS` | `admin_only` | management-surface access: `admin_only`, `private`, or `public` |

and the surrounding prose says "the management surface follows its tri-state setting". An operator reads that as *the* gate on `/management/*`. It is not: when `[management.endpoints]` sets per-endpoint levels — which the shipped quickstart config does, for every endpoint — the RBAC key has **no observable effect in either direction**.

That is a security-relevant surprise. Someone who sets `FERROEHR__AUTHZ__RBAC__MANAGEMENT_ACCESS=admin_only` to lock the surface down has not locked it down.

## Reproduction

`ghcr.io/rubentalstra/ferroehr:develop` (`git_sha 1785c2ef9ac32f3636d81936829acd4304c3bc39`), quickstart Compose + the Keycloak overlay, `FERROEHR__AUTHZ__RBAC__ENABLED=true`, bearer principals from the demo realm with distinct realm roles.

Effective configuration, read back from the server:

```
[management.endpoints]  info=private metrics=private prometheus=public env=admin_only loggers=admin_only
[authz.rbac]            enabled=true management_access=admin_only
```

Observed (HTTP status per principal per endpoint):

```
USER           info      metrics   prometheus  env       loggers
clinician      200       200       200         403       403        (roles: USER)
adminuser      200       200       200         200       200        (roles: ADMIN,USER)
readonlyuser   200       200       200         403       403        (roles: READONLY,USER)
norole         200       200       200         403       403        (roles: none)
(anonymous)    401       401       200         401       401
```

`management_access = admin_only` did not stop `norole` — a principal with no roles at all — from reading `/management/info` and `/management/metrics`.

Flipping the same key the other way changes nothing either:

```
$ FERROEHR__AUTHZ__RBAC__MANAGEMENT_ACCESS=public   (confirmed via /management/env)
--- anonymous, with [management.endpoints] info=private:
  info      -> 401
  metrics   -> 401
  env       -> 401
  loggers   -> 401
```

So in the shipped posture the key is inert in both directions.

## What is probably true, and what needs saying

`website/book/src/installation/configuration.md` documents the precedence one level down: `[management] access_default` — "Global default access level (**a per-endpoint level wins**)". The observed behaviour is consistent with `authz.rbac.management_access` being a *default* of the same kind, overridden by any per-endpoint level. If so, the defect is that two differently-named tri-state knobs govern one surface, are documented in two different chapters, and neither chapter mentions the other or states which wins.

## Acceptance criteria
- [ ] The precedence between `authz.rbac.management_access`, `management.access_default` and `management.endpoints.<name>` is stated in both chapters that mention any of them
- [ ] If the RBAC key is genuinely redundant with `management.access_default`, it is deprecated rather than left as a second spelling
- [ ] The matrix above is pinned by a test, so a future change to the precedence is caught

### LINKED ISSUE #2192: ci(chart): the chart advertises a postgres image it never deploys, dragging that image's CVEs onto our listing
The Artifact Hub listing reports security findings against `ghcr.io/rubentalstra/ferroehr-postgres`. **The chart does not deploy a database.** It only *declares* that image, and the declaration alone is what pulls those findings onto our package.

## What is actually true

The chart has no database in it, and this was verified rather than assumed:

- `deploy/helm/ferroehr/templates/` — thirteen templates, none of them a database (deployment, service, configmap, secret, ingress, hpa, pdb, networkpolicy, serviceaccount, servicemonitor, migration-job, NOTES, _helpers)
- no `charts/` directory, no `dependencies:` in `Chart.yaml` — no subchart
- `values.yaml` `database:` takes an **external** DSN via `existingSecret`, with the inline `url` marked DEV/TEST ONLY

So the architecture already matches the position the docs take: the database lives outside the cluster, because a cluster that goes away must not take the clinical record with it.

The one place `ferroehr-postgres` appears is `Chart.yaml`:

```yaml
artifacthub.io/images: |
  - name: ferroehr
    image: ghcr.io/rubentalstra/ferroehr:…
  - name: ferroehr-postgres          # ← the chart never deploys this
    image: ghcr.io/rubentalstra/ferroehr-postgres:…
```

Artifact Hub scans every image in that annotation and attributes the results to the package. So we are being reported for the CVE surface of an image **no chart user ever runs** — an upstream `postgres` base with everything a full database image carries, versus our distroless server.

`ferroehr-postgres` is not junk: it is the quickstart Compose database, preconfigured with the roles, schemas and extensions, and that is a genuinely useful dev convenience. It simply has nothing to do with the chart.

## Why this is worth a P1 rather than a cosmetic fix

Three consequences, in order of seriousness:

1. **The vulnerability report is wrong about what we ship.** Anyone evaluating the chart sees findings for software the chart does not install. That is a false negative in the other direction too — real findings in the actual server image compete for attention with noise.
2. **It misrepresents the architecture.** An image list is read as "what this chart runs". Advertising a database says we ship one, which contradicts both the chart and the hardening chapter.
3. It is the difference between an accurate listing and a permanently red one, since the upstream postgres base will always carry more CVEs than a distroless binary.

## The check to add, not just the deletion

Removing the entry is one line. The thing worth building is the guard: **every image in `artifacthub.io/images` must be one the chart's rendered output actually references, and every image the rendered output references must be listed.** Both directions — a missing entry means an unscanned image, an extra entry means this. `deploy/helm/validate.sh` already renders every overlay, so it has the material.

## Acceptance criteria
- [ ] `ferroehr-postgres` removed from the chart's `artifacthub.io/images`
- [ ] A guard in `deploy/helm/validate.sh` reconciling that annotation against the images the rendered manifests actually reference, in **both** directions, mutation-proven each way
- [ ] The next published chart's Artifact Hub listing shows security findings only for images the chart deploys — confirmed on the live listing, not inferred
- [ ] The compose quickstart's use of `ferroehr-postgres` left untouched and its purpose stated where the image is defined, so nobody deletes it thinking it was chart-only
- [ ] The hardening chapter says plainly that the database is external and the chart never provisions one, if it does not already

### LINKED ISSUE #2193: feat(chart): the Helm chart cannot deploy the admin console, though it advertises it
Split from #2192. The chart's `artifacthub.io/images` lists `ferroehr-admin-ui`, and the owner's intent is that the console is an optional companion an operator can run. **The chart cannot deploy it**, so the listing advertises something the chart does not provide.

Verified rather than assumed:

- no `adminUi`/`admin_ui` key anywhere in `values.yaml`
- no admin-console template — the thirteen templates are the server's, and there is exactly one `deployment.yaml`
- the only `admin` reference in the templates is `NOTES.txt` printing `config.admin.enabled`, which is the admin **API** (a server feature), not the console
- all three CI overlays render **zero** references to the image

```
$ for v in default all-features basic-auth; do
    helm template t deploy/helm/ferroehr -f deploy/helm/ci/$v-values.yaml | grep -c ferroehr-admin-ui
  done
0
0
0
$ helm template t deploy/helm/ferroehr -f deploy/helm/ci/default-values.yaml | grep -oE 'image: "[^"]*"' | sort -u
image: "ghcr.io/rubentalstra/ferroehr:3.17.3"
```

The image and its Compose profile are real and work; only the chart is missing.

## Why this matters beyond tidiness

Artifact Hub scans every image in that annotation and attributes the findings to the package, so the console's vulnerability surface is already reported against the chart. That is *reasonable* if the chart can deploy it and misleading if it cannot — which is the same defect that #2192 fixed for `ferroehr-postgres`. The difference is the intent: postgres should never be a chart component (the database is external, deliberately), whereas the console **should** be one and is not yet. So the resolution differs: delete there, build here.

The entry stays listed in the meantime, with a comment at the annotation saying it advertises intent rather than behaviour, so the discrepancy is visible to the next reader instead of looking like an oversight.

## What "optional companion" has to mean in a chart

An `adminUi.enabled: false` block that, when switched on, produces the console's own Deployment, Service, ServiceAccount and optional Ingress — with the properties the server's already has and which are easy to lose on a second workload:

- the same pod-security posture (`runAsNonRoot`, numeric UID, `readOnlyRootFilesystem`, dropped capabilities, seccomp `RuntimeDefault`)
- a NetworkPolicy that lets it reach the CDR and nothing else — it is a REST client of the CDR by mandate, so its egress is narrow and worth enforcing rather than documenting
- its session/OIDC configuration through the same secret-as-file route the server uses, never env-borne
- `config.files`-style mounting for anything file-shaped it needs
- off by default, and off leaving no residue

## Acceptance criteria
- [ ] `adminUi.enabled` deploys the console, and defaults to off with zero rendered objects when off
- [ ] The console's pod-security posture matches the server's, asserted by the chart's own security gate rather than by review
- [ ] A NetworkPolicy restricting its egress to the CDR, with the rule tested rather than only rendered
- [ ] Golden renders for the enabled case, and `deploy/helm/ci/` gains an overlay that boots it
- [ ] Verified on a live cluster: the console comes up, reaches the CDR over ITS-REST, and a browser session works — per #2164
- [ ] The `artifacthub.io/images` comment added in #2192 removed once the entry describes real chart behaviour

### LINKED ISSUE #2194: refactor(chart): rebuild the Kubernetes deployment against the official guidance
Working plan: `docs/plans/k8s-chart-architecture.md` (deleted in the implementing PR). Every rule below is from the official Kubernetes or Helm docs and cited there; this issue records the **delta** — where our chart contradicts them and what that costs.

## 1. The liveness and readiness probes are the same command

The server's liveness and readiness probes both run `ferroehr healthcheck`, differing only in timing. Readiness must consult the database or it is meaningless — which means a database blip **restarts every pod** instead of removing them from the Service.

That is the documented anti-pattern verbatim ([Probes](https://kubernetes.io/docs/concepts/workloads/pods/probes/)):

> Incorrect implementation of liveness probes can lead to cascading failures. This results in restarting of container under high load; failed client requests as your application became less scalable; and increased workload on remaining pods due to some failed pods.

and the division of labour is explicit:

> The liveness probe passes when the app itself is healthy, but the readiness probe additionally checks that each required back-end service is available.

**Liveness must be process-local. Readiness owns dependencies. A startup probe protects the cold start** (migrations, template cache) from a liveness probe that would otherwise kill it mid-boot.

## 2. The PDB omits the documented recommendation

[Disruptions](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/):

> It is recommended to set `AlwaysAllow` Unhealthy Pod Eviction Policy to your PodDisruptionBudgets to support eviction of misbehaving applications during a node drain. The default behavior is to wait for the application pods to become healthy before the drain can proceed.

We never set `unhealthyPodEvictionPolicy`. Combined with our `maxUnavailable: 0` posture, a node drain waits for unhealthy pods to become healthy — and if they are unhealthy *because* the node is draining, that wait never ends. The operator sees a hung drain with no explanation.

## 3. Restricted-profile compliance is by luck, not by check

Our pods satisfy the [restricted profile](https://kubernetes.io/docs/concepts/security/pod-security-standards/), but the profile is enforced by Pod Security Admission through **namespace labels**. A chart producing compliant pods into an unlabelled namespace is compliant by accident: nothing in the chart or the docs tells an operator to set `pod-security.kubernetes.io/enforce: restricted`, and nothing fails if they do not.

## 4. An immutability trap, found the hard way

A Deployment's `spec.selector.matchLabels` is **immutable**. Adding `component` to the shared `selectorLabels` helper — the obvious way to separate server pods from console pods — breaks `helm upgrade` on every existing release, with an error most operators read as a chart bug. I started to do exactly this while building #2193 and caught it before it landed.

`component` belongs on object and pod-template labels only. New workloads build their own selector from name+instance+component from the start.

## 5. Configuration hygiene

From [Configuration Best Practices](https://kubernetes.io/docs/concepts/configuration/overview/): *"Don't specify default values unnecessarily"* — several templates set fields to their API defaults, noise in every rendered manifest and every golden diff. `kubernetes.io/description` is called out as the most useful annotation and we use it nowhere. Every `apiVersion` wants an audit against current stable.

## Acceptance criteria
- [ ] Liveness is process-local and provably cannot fail on a dependency outage — demonstrated by stopping the database and observing pods go **not-ready without restarting**
- [ ] A startup probe covers the cold start, sized against a measured boot rather than a guess
- [ ] `unhealthyPodEvictionPolicy: AlwaysAllow` by default, with the trade stated where it is set
- [ ] `validate.sh` asserts every rendered pod would pass the **restricted** profile — control by control, not a spot check — and the namespace label is documented as the prerequisite it is
- [ ] `component` labels on objects and pod templates, **never** in a selector; a test that the server's `selector.matchLabels` is byte-identical to the previous release, since that is the upgrade-breaking field
- [ ] Default-valued fields removed; `kubernetes.io/description` on the objects an operator meets first
- [ ] Golden renders regenerated and the whole thing installed on a live cluster, not merely rendered

### LINKED ISSUE #2197: fix(multimedia): EHR_STATUS and FOLDER blobs are externalized but can never be read back
Externalization is applied to **every versioned object**; re-inlining is wired to **compositions only**. So a `DV_MULTIMEDIA` committed inside an `EHR_STATUS` (or a `FOLDER`) leaves the database and there is no API that returns it again.

Found while verifying #2168/#2169 on a live compose stack, not by reading code.

## Reproduction

Quickstart Compose + `--profile s3`, server built from `fix/observability-push-and-chart-images`, multimedia enabled via the (now working) shell exports.

Commit an EHR whose `EHR_STATUS.other_details` carries a 400 KiB `DV_MULTIMEDIA`:

```
POST /ehr   201
```

Read the status back — the blob was externalized, exactly as intended:

```
has inline data : False
uri             : s3://openehr-multimedia/6614295bbad8b3b44c9eafc3ac6def601c42630c24654395c8512d8854f2056f
size            : 409600
integrity alg   : SHA-256
```

and the object is really in the bucket:

```
$ curl -s "http://localhost:18333/openehr-multimedia?list-type=2" | grep -o '<Key>[^<]*</Key>'
<Key>6614295bbad8b3b44c9eafc3ac6def601c42630c24654395c8512d8854f2056f</Key>
```

Now ask for it back:

```
$ curl -u … "…/ehr/$EHR/ehr_status?expand_multimedia=true"
200, 1218 bytes — no `data` key; the compact s3:// reference again
```

The 400 KiB are in the object store and **no API call returns them**.

## Why

- `offload` is called in the generic versioning path (`versioning/change.rs`), which every versioned object commits through — composition, EHR_STATUS, folder alike.
- `expand_multimedia` is called from exactly one handler, `GET /ehr/{ehr_id}/composition/{uid_based_id}` (`api/ehr/composition.rs`), and the OpenAPI parameter is declared only there — its own description says "so the served COMPOSITION is byte-identical".

`?expand_multimedia=true` on the EHR_STATUS read is therefore an undeclared parameter and is ignored, which is ordinary HTTP behaviour and not itself the bug. The bug is the asymmetry: the write path externalizes content the read path has no way to restore.

This is a sibling of #2171 (disabling the integration stranded already-offloaded blobs) and the same failure class — content the server moved becoming unreachable through the API — but a distinct cause: that one was a config gate, this one is a missing seam.

## The options
1. **Symmetry**: wire `expand_multimedia` into the EHR_STATUS and FOLDER (directory) reads and declare the parameter there. Content is restorable wherever it was offloaded.
2. **Restrict the write**: offload ONLY from compositions, leaving EHR_STATUS/FOLDER media inline. Narrower, and it changes storage behaviour for records already committed.

(1) matches what the offload path already does and is the smaller change to reason about; (2) would need a decision about blobs already externalized from non-composition objects.

## Acceptance criteria
- [ ] A `DV_MULTIMEDIA` committed in an `EHR_STATUS` is retrievable in full through the API, or is never externalized in the first place
- [ ] The same settled for `FOLDER`/directory content
- [ ] Whichever way it goes, the OpenAPI parameter declarations match the handlers that honour it
- [ ] Live evidence on a real stack, byte-identical to what was committed

rubentalstra:
Closes #2173. Closes #2192. Closes #2193. Closes #2170. Closes #2194. Closes #2176. Closes #2177. Closes #2169. Closes #2168. Closes #2171. Closes #2197. Closes #2066. Closes #2165.

**BREAKING:** the chart requires Kubernetes **1.36+**, `probes.exec` is gone, and two dead authorization keys are removed. Chart 5.0.1 → **6.0.0**.

Verified on a live cluster (Docker Desktop, Kubernetes **v1.36.1**, containerd 2.3.1), not rendered.

## The Kubernetes floor is 1.36, and user namespaces are what it buys

Every pod the chart deploys now runs in its own **user namespace** (`hostUsers: false`). Read off the running container's runtime spec:

```
uidMappings: [{containerID: 0, hostID: 1067319296, size: 65536}]
gidMappings: [{containerID: 0, hostID: 1067319296, size: 65536}]
```

Container root is host UID 1,067,319,296 — a user that owns nothing. Capabilities held inside the namespace do not apply outside it. This is the strongest per-pod isolation Kubernetes offers a workload holding PHI, and `hostUsers` went **stable in 1.36** (KEP-127), which is the whole reason the floor moves there.

The floor is a real compatibility floor, not a support opinion. `Chart.yaml` now carries the field-to-KEP table that is its evidence, taken from each `keps/<sig>/<kep>/kep.yaml` rather than a docs page — a docs page *drops* the version note once a feature goes GA, so it cannot answer the question:

| field | KEP | stable |
|---|---|---|
| `hostUsers: false` | 127 | **v1.36** ← sets the floor |
| `supplementalGroupsPolicy` | 3619 | v1.35 |
| `preStop.sleep` | 3960 | v1.34 |
| `Service.trafficDistribution` | 4444 | v1.33 |
| `PDB.unhealthyPodEvictionPolicy` | 3017 | v1.31 |
| `Job.podFailurePolicy` | 3329 | v1.31 |

Nothing in the chart is version-gated any more. The cost is stated rather than glossed: 1.36 is the **top** of the supported window, so a patched 1.34 or 1.35 cluster is refused. A gate is a silent absence; a floor is a loud refusal, and for this workload the loud one is correct.

**AppArmor is deliberately NOT enabled with it.** Measured on the same node: a pod requesting `appArmorProfile: RuntimeDefault` where the host has no AppArmor is *rejected outright* — `Cannot enforce AppArmor: AppArmor is not enabled on the host`, status `AppArmor`, container never started. It stays opt-in with its prerequisite written down. Rendering would never have found that.

## The liveness proof, on the cluster

Database scaled to zero, in a namespace enforcing `restricted`:

```
ferroehr-84b9c876b8-sgq9j   READY=false  RESTARTS=0
ferroehr-84b9c876b8-tdhpm   READY=false  RESTARTS=0

readiness 503  {"status":"DOWN","db":{"status":"DOWN"},"migrations":{"status":"DOWN"}}
liveness  200
endpoints {"ready":false,"serving":false}
```

Not-ready **without restarting**, withdrawn from the Service, liveness untouched. Database back → `readiness UP`, `POST /ehr` → **201**, still zero restarts.

Enforcement was live and independently proved itself: with the namespace labelled `restricted`, the stock postgres *fixture* was refused (`allowPrivilegeEscalation != false`) while every FerroEHR pod was admitted.

## Availability defaults that were missing

- **Replicas now spread across nodes.** Two replicas with nothing separating them can share one node, making a single node failure a total outage. A soft `maxSkew: 1` over `kubernetes.io/hostname` (`ScheduleAnyway`, so single-node clusters still schedule), with `matchLabelKeys: [pod-template-hash]` so a rollout is not counted as skew.
- **`preStop.sleep`** covers the lame-duck window: endpoint removal and SIGTERM happen concurrently, so without it a terminating pod still receives requests. Native `sleep`, because the image ships no shell for an `exec` hook.
- **The migration Job ignores disruption-caused failures**, so a node drain during a release cannot burn `backoffLimit` and fail the upgrade with no migration error in any log.

## Two gates that were weaker than they looked

**The console was never gated.** `validate.sh` rendered three overlays and *none* enabled `adminUi` — so the per-container restricted-profile check, which exists precisely because a second workload is where a posture gets lost, had never once seen the console. It is now a fourth case with its own golden.

**The first spread assertion was wrong, and mutation caught it.** It read a missing `replicas` as the API default of 1, which exempted exactly the autoscaled workload that most needs spreading. Both new gates — isolation parity across workloads, and multi-replica spread — are mutation-proven: each fails naming the offending object, and the unmutated render passes.

## Two documented gates on the management surface, neither of them wired (#2177)

`authz.rbac.management_access` is presented by the security chapter as *the* gate on `/management/*`. Nothing produces the operation class it switches on, so it had no effect in either direction — someone setting `admin_only` to lock the surface down had locked nothing. `management.access_default` is the same defect one level down: documented as the fallback a per-endpoint level overrides, never read by the router that resolves those levels.

Both are **removed**, not deprecated — there is no behaviour to preserve, and a second spelling of a setting is how this happened. `[management.endpoints]` is the single authority; an endpoint that names no level is `off` and is not mounted.

Pinned by a test the old suite could not have caught: the per-level tests set one endpoint at a time, so a guard applying the most permissive configured level to every route would have passed all of them. The new case runs four endpoints at four levels in one render and asserts every principal against every one.

Four documentation sites carried the wrong story. The operations chapter additionally claimed every endpoint defaults to `admin_only` when the real default is `off` — the `docs-claims` guard caught that page after the first three were fixed by hand.

## The development Keycloak realm could not mint an acceptable token (#2176)

The realm declared no audience mapper, and Keycloak puts a client in `aud` only when a mapper says so. The development configuration requires `audiences = ["ferroehr"]`, so every token it issued was refused `401 InvalidAudience` — not a misconfiguration an operator could hit, an impossibility. The server is right on both halves (RFC 7519 §4.1.3); the realm was wrong. Verified after the fix: `aud: ferroehr`, `azp: ferroehr`, `realm roles: [ADMIN]`.

The realm was also carrying the identity of the deployment it had been exported from — a hardcoded tenant claim naming a foreign UUID, a foreign client scope, a default role named after another product. Nothing referenced any of them.

## Multimedia: the documented recipe did nothing, twice (#2169, #2168)

`FERROEHR__MULTIMEDIA__*` exported in a shell never reached the container — every *other* documented tunable is a pass-through and these were not, so the stack came up healthy with multimedia silently off. The whole set is now declared with **no value**, the Compose form that forwards a variable when set and *removes it when unset*. That form over `${VAR:-}` deliberately: an empty `endpoint` is a boot refusal, not a synonym for unset, so the convenient spelling would have turned a silent no-op into a stack that refuses to start.

And the gateway ships with **no bucket**, while an S3 write into a missing bucket answers `403 AccessDenied` rather than `404 NoSuchBucket` — so the failure presented as a credentials problem on a gateway that needs none. `seaweedfs-init` now performs the CreateBucket once the gateway is healthy; it is idempotent by AWS's own definition, reads the same bucket variable the server does, and carries the compose isolation floor.

## Disabling multimedia stranded every blob it had offloaded (#2171)

`?expand_multimedia=true` against an already-externalized record was **silently ignored** once `multimedia.enabled` went back to `false`: `200`, the compact `s3://` reference, no error. The caller asked for clinical bytes and got a URI.

`enabled` now governs **new offloads** only — the fetch-verify-reinline path stays available while an endpoint is configured, so content this server externalized stays readable. With no store reachable at all the expansion request **fails** rather than answering with the reference. A read-back-only store never fails the boot, deliberately: turning an integration off must not become a way to break startup.

## Verified on a live compose stack

The multimedia fixes were driven end to end against a real stack (server built from this branch, SeaweedFS gateway, quickstart Compose), which is also where the remaining finding below came from:

- **Exports reach the server.** `/management/env` reports `enabled: true`, `endpoint: http://seaweedfs:8333`, `allow_http: true` — from shell exports alone, which previously reached nothing.
- **The bucket exists without a manual step.** `seaweedfs-init` exits 0; the object write the issue recorded as `403 AccessDenied` now returns `200`.
- **A 400 KiB `DV_MULTIMEDIA` offloads correctly.** `POST /ehr` → `201`; the stored record holds no inline `data`, a `s3://openehr-multimedia/6614295b…` uri, `size: 409600` and `SHA-256` as its integrity algorithm; the object is in the bucket under that key.

The initializer ended up reusing the gateway's own image — it already ships curl — so the profile adds no second image to pull, pin or scan, and initializer and gateway cannot drift.

**Honest scope:** #2171's runtime wiring is covered by unit and integration tests, not by the live run above — the image used predates that commit, and rebuilding it was not worth another hour to re-observe what the tests already assert.

## A defect the live run exposed, and fixed (#2197)

The 400 KiB above went in through an `EHR_STATUS` — and could not come back out. Externalization lives in the generic versioning path, so a `DV_MULTIMEDIA` leaves the database from a COMPOSITION, an EHR_STATUS and a FOLDER alike; re-inlining was wired to the COMPOSITION read alone. `?expand_multimedia=true` on the EHR_STATUS read was an undeclared parameter and was ignored — correct HTTP, and precisely the silence that hid it.

Fixed by **symmetry**, not by narrowing the write. Restricting offload to compositions was the smaller diff, but it would leave every blob already externalized from a non-composition object exactly as unreachable as it is today: the choice is between making old content readable and making new content stay put, and only one of those helps a deployment that already has the data.

All nine reads that can serve externalized content now route through one seam, and the served OpenAPI declares the parameter on exactly those nine operations — pinned by a test, because a declaration list is the half a document can check and it catches the direction that actually bit.

## Object-level default-deny, as one setting (#2066)

An EHR carrying no `ACCESS_CONTROL_SETTINGS` admitted any caller the coarse layers had already let through — a defensible posture, honestly documented. What was not defensible: the *safer* posture was reachable only by authoring a settings object **per EHR, through a CONTRIBUTION commit per record**, including for EHRs created by clients.

`authz.rbac.ehr_access_default = "restricted"` says it once. The default stays `open`, because this changes who can read **existing** records — a deployment's decision, not an upgrade's. Explicit per-EHR settings win over it in both directions.

The admin carve-out is the arguable part, so the reason is in the code: a plain deny leaves a setting-less EHR unreachable by everyone, including the operator who would author the settings that open it. Framing is the OWASP IDOR cheat sheet's — `ehr_id` is a UUIDv7, and an unguessable id is not an access control.

## Every optional integration now says how to enable it on Kubernetes (#2165)

Audited all nine by following the book alone. One finding dominated: **not one integration chapter mentioned Helm** — change events, FHIR, terminology, Subject Proxy, S3 multimedia, the console, observability, signing, tenancy all documented only the env/Compose path, while the book presents Kubernetes as the production path. The capability was never missing (the `config` passthrough has always reached every key), which made it the worst kind of gap: present and invisible. The console was sharpest — its chapter never mentioned `adminUi`, so the workload this release taught the chart to deploy had no documented way to reach it.

Each now carries a values snippet, how to turn it off, and the prerequisites stated **before** the reader meets them — which matters most for the three that fail *silently*: an OTLP exporter blocked by the default-deny egress policy (no traces, no error), a Subject Proxy system behind the same policy (timeouts), and a tenancy claim the IdP never issues (runs unscoped against the reserved default, so it looks like tenancy is doing nothing).

## Greenfield documentation

The reference material had accumulated version history — which floor preceded this one, which chart version first carried a signature. None of it helps someone deploying today and every line is a claim that must stay true forever. `CHANGELOG.md` keeps the history; that is its job and the only place it belongs.

## Still open, named rather than implied

`_helpers.tpl`, `deployment.yaml` and `values.yaml` have had no structural pass, and the console has not been driven through a browser on the cluster (#2164).


rubentalstra:
Two commits added since the description above.

## A silent-nothing defect the new rules found immediately

The chart declares `kubeVersion: ">=1.25.0-0"` and rendered
`PodDisruptionBudget.unhealthyPodEvictionPolicy` unconditionally. KEP-3017
(`keps/sig-apps/3017-pod-healthy-policy-for-pdb/kep.yaml`) records **alpha
v1.26, beta v1.27, stable v1.31**. On 1.25 the API server prunes the field; on
1.26 it sits behind an alpha gate that is off by default. Either way the chart
installed cleanly while advertising the drain-safety property this PR's
description claims it has.

Gated at 1.27, where it takes effect by default. Verified: `--kube-version
1.25.0` omits it, `1.27.0` and `1.31.0` render it, default render unchanged so
no golden churn.

**Why I missed it, and why a reviewer would too:** once a feature goes GA the
version note is *removed* from the API reference page, so the current page for
a field added in 1.31 reads exactly like one for a field present since 1.0.
The docs site cannot answer this question. The KEP's own `kep.yaml` can, and
that lookup is now the documented procedure.

## The rules that were missing

`.claude/rules/kubernetes-helm.md` — the repo had 552 lines of skill for
*testing* a chart and nothing at all for *authoring* one, which is why this
rewrite kept guessing. Eight rules, each named after a defect this work
actually produced: version-gating, immutable `selector.matchLabels`, gates
that grep and pass vacuously, diagnosing off a non-default branch, probe roles
driving *server* design, the no-database principle and its Artifact Hub
consequence, PSA as the operator's job. The enforcement register marks the two
properties no tool can check rather than implying they are covered.

Also adds a `k8s-reviewer` agent, and fixes something real in the four
existing reviewer agents: each declares `tools: Read, Grep, Glob, Bash` and
calls itself read-only, but the runtime grants Write and Edit on top — so
"read-only" was a claim the definition did not enforce. `disallowedTools`
("removed from inherited or specified list") is the documented mechanism and
is now applied. Their declared tool lists never included Write, so this
changes no intent; it makes the existing intent true.

PR #2205, opened 2026-08-10
test(fuzz): the fuzzing lane never fuzzed anything

### LINKED ISSUE #2204: test(fuzz): the fuzzing lane has been dead — every target fails to build, and nothing on the PR path would notice
The `Fuzz` workflow has failed on every scheduled run. Not a flaky target and not a crash finding: **all six harnesses fail to compile**, so no fuzzing has happened at all.

```
$ gh run list --workflow fuzz.yml --limit 3
failure  2026-08-10T03:20  develop  Fuzz
failure  2026-08-09T03:17  develop  Fuzz
failure  2026-08-08T03:12  develop  Fuzz
```

All six jobs, same error:

```
error: sanitizer is incompatible with statically linked libc,
       disable it using `-C target-feature=-crt-static`
error[E0463]: can't find crate for `core`
error: could not compile `libc` (lib) due to 2 previous errors
Error: failed to build fuzz script
```

## Root cause, from the failing command itself

The build line in the log ends:

```
"cargo" "build" "--manifest-path" ".../fuzz/Cargo.toml" \
  "--target" "x86_64-unknown-linux-musl" "--release" ...
```

**`cargo-fuzz` defaults the build target to the triple IT was built for, not the runner's.** The workflow installs it through `taiki-e/install-action`, which resolves cargo-fuzz via its cargo-binstall fallback to the upstream **musl** asset — the workflow's own comment notes this, but not that it changes the *build target*. musl links libc statically, and a static libc cannot carry `-Zsanitizer=address`. So every build died before a single input was executed.

The fix is to name the triple explicitly (`--target x86_64-unknown-linux-gnu`), which is the documented way out and keeps the choice visible rather than dependent on which asset the installer picked.

## The structural half, which matters more

The lane is **scheduled-only** (`cron` + `workflow_dispatch`). Nothing on the pull-request path ever compiled the harnesses, so:

- a build break like this one surfaces only at 02:41 UTC, into a workflow nobody is watching;
- a fuzzed crate renaming a function breaks a harness silently, which is the *most likely* way a target rots;
- there is no signal at review time that a change to `openehr-its`, `openehr-query`, `openehr-rm` or `openehr-adl` broke the thing that reads their bytes.

A scheduled-only quality lane decays to zero and reports success by silence. That is the same failure class the deployment-probe work (#2178) exists to end, in a different corner.

## Why this matters beyond a red badge

`fuzz/README.md` records what this lane is for, and it is not hypothetical: a missing nesting bound in the XML reader recursed the generated `FromXml` impls off the stack, and a Rust stack overflow **aborts** rather than unwinding — so the `tower-http` catch-panic layer that renders this server's clean `500` could not intercept it. One malformed request would have taken the process down for every caller.

That class is reachable from unauthenticated request bytes. The instrument that catches it has been off.

## Acceptance criteria
- [ ] Every target builds and runs in CI, with the target triple named explicitly and the reason recorded where someone would remove it
- [ ] A PR-path job that **compiles** the harnesses whenever `fuzz/**` or a fuzzed crate changes — fast enough that nobody wants it disabled
- [ ] The standing discipline written down as a rule file (harness purity, a crash is fixed in the crate never in the harness, coverage only ratchets up)
- [ ] Coverage extended beyond the six: the identifier readers reached from the `{version_uid}` PATH parameter are parsed before any body, negotiation or authorization and are currently unfuzzed
- [ ] A green scheduled run recorded here, so "it works" is an observation rather than a claim

rubentalstra:
Closes #2204

The `Fuzz` workflow failed on every scheduled run, and not on a finding — **all six harnesses failed to compile**, so no input was ever executed.

## What was wrong

```
error: sanitizer is incompatible with statically linked libc,
       disable it using `-C target-feature=-crt-static`
```

`cargo-fuzz` defaults its build target to the triple **it** was built for, not the runner's. CI installs it through `install-action`'s binstall fallback, which resolves the upstream **musl** asset, so every build silently went to `x86_64-unknown-linux-musl` — whose static libc cannot carry `-Zsanitizer=address`. The workflow already noted that the installer picks musl; what it did not note is that this decides the *build target*.

Naming `x86_64-unknown-linux-gnu` explicitly fixes it, and keeps the choice visible instead of dependent on which asset an installer resolved.

## The half that matters more

The lane was **scheduled-only**. Nothing on the pull-request path ever compiled a harness, which is why a total build break sat unnoticed across nights: it reported into a 02:41 UTC workflow with no reader. And a build break is not the exotic failure here — a fuzzed crate renaming a function is by far the likeliest way a target rots, and the campaign is the worst possible place to learn it.

There is now a `build` job on the PR path, triggered by changes to `fuzz/**` or any fuzzed crate. It compiles every harness and nothing else, so it is fast enough that nobody will want it turned off.

## A target for the identifier readers

They had none, and they are the earliest attacker-reachable parser in the server: every versioned read takes `{version_uid}` from the URL **path** and hands it to `ObjectVersionId::from_str` before a body is read, a content type is negotiated, or authorization runs.

`OBJECT_VERSION_ID` is a composite (BASE `base_types` master05 §Composite Identifiers) — `object_id::creating_system_id::version_tree_id`, with the third part itself dotted. Two separator kinds and a nested grammar is the exact shape where a reader mis-slices and still returns `Ok`, which no panic-only harness would catch. So the target walks the parsed value's accessors and then requires the composite's own contract to hold:

```rust
assert_eq!(
    ObjectVersionId::compose(&object_id, &system, &tree),
    ovid,
    "an OBJECT_VERSION_ID recomposed from its own parts must equal itself"
);
```

Seeded with seven literals covering the trunk, branch and malformed-separator shapes.

## The standing discipline, written down

`.claude/rules/fuzzing.md` records what this lane is and the traps it has actually hit — the `--target` one above, `fuzz/` being its own nightly workspace on purpose, and why a scheduled-only quality lane decays to zero. The load-bearing rule is the one that mirrors CNF triage: **a crash is fixed in the crate, never in the harness** — never widen a bound, delete a target, or disable the sanitizer to go green, and every fixed crash gets a regression test in the owning crate from the recorded artifact. The root `CLAUDE.md` carries the short form.

That rule is not abstract. The defect that motivated this lane was a missing nesting bound in the XML reader recursing the generated `FromXml` impls off the stack — and a Rust stack overflow **aborts** rather than unwinding, so the `tower-http` catch-panic layer that renders this server's clean `500` could not have intercepted it. One malformed request would have taken the process down for every caller. That class is reachable from unauthenticated request bytes, and the instrument that catches it has been off.

## Verification

- `cargo check --bin identifiers` in the fuzz workspace — clean
- `fuzz/seeds.sh` — `identifiers: 7 written`
- `zizmor --min-severity=low .github/workflows/fuzz.yml` — no findings
- The `build` job is what proves the rest, on this PR


rubentalstra:
Both halves verified first-hand.

**The build, in CI, on this PR** — the same step that failed every night:

```
harnesses build / Build every harness   completed   success
```

**The new target, locally**, on `aarch64-apple-darwin` with the sanitizer on:

```
Done 17680372 runs in 91 second(s)
```

17.6M inputs through the identifier readers with no crash, no hang, and the composite contract holding throughout — an `OBJECT_VERSION_ID` recomposed from the three parts its own reader produced equalled itself every time. That is the assertion worth having: a panic-only harness would have passed on a reader that mis-sliced a separator and still returned `Ok`.

Also applied CodeRabbit's finding, which was correct: `openehr-base` is a direct dependency of the harnesses and the pull-request paths filter omitted it, so a parser change there could have broken the build with nothing to notice. The path is added — and because the comment claiming that list mirrors `fuzz/Cargo.toml` would rot the same silent way, the build job now compares the two sets and fails with a diff. Proven by deleting the line again and watching the check refuse.

Issue #2211, opened 2026-08-10
ci(release): the crates.io publish split the lockstep set — six crates are at 0.0.15, two are stranded at 0.0.10


rubentalstra:
The `publish-crates` lane failed part-way through, and it failed **after** uploading. The result is a published state that the rules say cannot exist.

Verified against crates.io just now:

| crate | latest published |
|---|---|
| `openehr-base` | 0.0.15 |
| `openehr-lang` | 0.0.15 |
| `openehr-term` | 0.0.15 |
| `openehr-query` | 0.0.15 |
| `openehr-rm` | 0.0.15 |
| `openehr-am` | 0.0.15 |
| **`openehr-its`** | **0.0.10** |
| **`openehr-adl`** | **0.0.10** |

## Why a split set is worse than a failed publish

`.claude/rules/crates-publishing.md` requires these eight to move in lockstep, and that is not stylistic: while the line is `0.0.x`, **cargo treats every `0.0.x` as its own compatibility set**, so `0.0.15` and `0.0.10` are incompatible versions, not "one slightly behind". Six crates now declare `0.0.15` internal requirements that resolve to nothing for the two that never made it. Anyone who tries to build against the just-published set gets an unresolvable graph — and the failure will look like their mistake.

## What the log says

```
Uploaded  openehr-rm v0.0.15 to registry `crates-io`
note: waiting for openehr-am v0.0.15 or openehr-rm v0.0.15 to be available at registry `crates-io`.
      2 remaining crates to be published
Published openehr-am v0.0.15 at registry `crates-io`
error: no packages ready to publish but 2 packages remain in plan with
       1 awaiting confirmation: openehr-adl v0.0.15 and openehr-its v0.0.15
note: this is an unexpected cargo internal error
note: cargo 1.96.1
```

Cargo itself calls this unexpected and asks for a bug report. The shape is its `--workspace` publish orchestration: it uploads in dependency order and waits for each upload to become visible in the index before the dependents go, and here it concluded nothing was publishable while still holding two packages in the plan — one of them "awaiting confirmation" of an index entry that had in fact just landed. So this is very likely a race in the index-availability wait, not a defect in our manifests: the two stranded crates are precisely the two deepest in the dependency graph, which are the ones that wait the longest.

Run: <https://github.com/rubentalstra/FerroEHR/actions/runs/31370233256/job/93397340008>

## The immediate repair

[...]

PR #2213, opened 2026-08-10
ci(crates): make the publish lane resumable, and refuse to report success on a split set

rubentalstra:
Refs #2211 — the P0. This is the lane fix; the registry repair rides on it.

## What happened

`cargo publish --workspace` uploaded six of the eight crates at `0.0.15`, then died calling it its own internal error, leaving `openehr-its` and `openehr-adl` at `0.0.10`. While the line is `0.0.x`, **cargo treats every `0.0.x` as its own compatibility set**, so this is not "two crates slightly behind" — the six published crates declare `0.0.15` internal requirements that resolve to nothing. Every consumer gets an unresolvable graph.

Then the obvious repair turned out to be impossible:

```
$ cargo publish --workspace --locked
error: crate openehr-am@0.0.15 already exists on crates.io index
```

`cargo publish --workspace` is **all-or-nothing at the start** (it refuses the whole run if any member version exists) and **non-atomic at the end** (a late upload can fail after earlier ones landed). Those two together mean a partial publish cannot be finished by re-running the thing that half-finished it. The workflow header actually claimed the opposite — that a re-run "retries the missing crates" — which is now corrected, because a comment that promises a recovery path that does not exist is worse than no comment.

## The fix

**Resumable**: publish per crate in dependency order, treating "already exists" as *done* rather than as failure. A re-run finishes the set.

**Verified**: the lane had no idea what it had published — it exited on cargo's status, and nobody would have discovered the split without reading the log. It now reads the registry back and fails if any of the eight is missing at the expected version, with the index's eventual consistency waited out rather than raced. Success now means the published set is whole, which is the only thing the lockstep rule actually cares about.

## One constraint, recorded because it nearly went wrong

The repair has to publish **the exact tree the six came from**. `develop` still pointed at that commit when this was written; had an unrelated `openehr-its` change landed first, a re-run would have published *different code* under the same version number, with no way to tell the two apart afterwards. This PR only touches the workflow, so packaged crate bytes are unchanged.

## Verification

- `yaml` parses; `zizmor --min-severity=low` reports no findings
- The failure mode is quoted from the real run: <https://github.com/rubentalstra/FerroEHR/actions/runs/31372359528>

### DIFF
--- a/.github/workflows/publish-crates.yml
+++ b/.github/workflows/publish-crates.yml
@@ -17,9 +17,12 @@
 #   - `cargo publish --workspace` resolves the inter-crate dependency order
 #     itself (Cargo ≥1.90) and skips the `publish = false` members (the
 #     app/tools crates), so the eight spec crates are exactly what ships.
-#     Publishing is non-atomic upstream: if a late upload fails, re-running
-#     the workflow retries the missing crates (already-published versions
-#     fail individually without harming the rest).
+#     Publishing is non-atomic upstream AND all-or-nothing at the start:
+#     `cargo publish --workspace` refuses the entire run when ANY member
+#     version already exists, so a partial publish could not be finished by
+#     re-running it. This lane therefore publishes per crate in dependency
+#     order, treats "already exists" as done, and verifies the whole set
+#     against the registry before reporting success.
 name: Publish crates
 
 on:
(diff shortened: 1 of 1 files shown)


rubentalstra:
Both findings were correct, and this merged before I had read them. Addressed in #2214.

**The `curl`/`jq` one is a real bug in what landed**, and a nasty one: under `pipefail`, an HTTP error body reaching `jq` kills the step, so a single transient `429` would have announced a split set that does not exist — in the one place whose entire job is telling the truth about whether the set is whole. A verification gate that cries wolf on a rate limit is worse than none, because the next genuine split reads as noise. The request is now guarded and a failed response counts as "not seen yet".

**The policy conflict was right to stop on, and the resolution is to fix the rule.** `.claude/rules/changelog.md` did mandate `cargo publish --workspace`, and the lane no longer does that — deliberately, because #2211 demonstrated first-hand that the command cannot finish a partial publish: all-or-nothing at the start, non-atomic at the end, so the retry that would repair a split set is refused outright. Keeping the rule in force would have meant a rule whose factual premise is disproved blocking the only design that recovers the failure it exists to prevent. Both rule files now describe the resumable per-crate lane and record why, with the reproduction.

`--workspace --dry-run` stays exactly where it was, and the rule now says so: nothing is uploaded, so all-or-nothing costs nothing there and every member is checked together. The change is to the upload, not the check.

Issue #2298, opened 2026-08-12
ci(licensing): FOSSA scans via the hosted git integration, so .fossa.yml never applies and the dashboard stays red


rubentalstra:
`.fossa.yml` and `.fossabot.yml` landed in #2290, written against the published schemas. Neither is being read, and the dashboard still shows six findings.

## Two separate causes, both real

**1. The displayed scan is of an old revision.** The dashboard reports `dbda82814` — the actionlint merge, which predates every piece of the licensing work. At that commit the 34 CNF scripts were still vendored, `reliability.md` still claimed CC-BY-SA 3.0 for the CKM models, and `.fossa.yml` did not exist.

**2. The config is not read by the scan that runs.** The project shows `Package manager: Git`, `Depth: Direct` — FOSSA's **hosted git integration**. `.fossa.yml` is the **CLI's** configuration file (<https://docs.fossa.com/docs/cli/references/files/fossa-yml>). The hosted scan never opens it, so the `targets.only: cargo`, the vendored-tree exclusions and the `customLicenseSearch` rules all currently do nothing.

## What a re-scan alone would and would not fix

| Finding | Clears without the CLI? |
|---|---|
| CC-BY-ND-3.0 | **Yes** — the 34 CNF scripts are deleted from the repository |
| AGPL-3.0-or-later | No — the file is still in git; only the *published crate* stopped shipping it |
| CC-BY-SA-3.0 | No — the openEHR specification text is in the tree, correctly |
| CC-BY-SA-4.0 | No — the CKM clinical models are in the tree, correctly |
| GPL-2.0 / LGPL-2.1 | No — the tri-licensed ISO 13606 test corpus is in the tree, correctly |

Five of the six are vendored third-party material that **should** be in the repository. They stop being findings only when something applies the exclusions — which means the CLI.

## The work

Add a CI job running `fossa analyze` from the FOSSA CLI, so the committed configuration is what governs. Requirements, all verifiable:

- Written against the official GitHub Actions integration reference, not from memory — the first `.fossa.yml` was written from assumption and was wrong until checked against the published schema.
- `uses:` pinned to a full commit SHA with the version in a trailing comment, `permissions: {}` at workflow level with the minimum per job, `persist-credentials: false` on checkout, no context value interpolated into a `run:` block. The `zizmor` job audits all four.
- Needs a `FOSSA_API_KEY` repository secret.
[...]

PR #2299, opened 2026-08-12
ci: run FOSSA through the CLI so the committed configuration governs

rubentalstra:
`.fossa.yml` landed with the licensing work and **nothing has ever read it.**

The project is scanned by the hosted GitHub integration (`Package manager: Git`, `Depth: Direct`), which analyses server-side with its own credentials and does not open that file — it is the CLI's (<https://docs.fossa.com/docs/cli/references/files/fossa-yml>). So the Cargo-only target, the vendored-tree exclusions and the `customLicenseSearch` rules have all been inert, and the dashboard has been reporting openEHR's specification text and a tri-licensed test corpus as though they were ours to license.

## Why analyse-only

`run-tests` is deliberately `false`. `fossa test` fails the build on a policy verdict, and this repository legitimately redistributes CC-BY-SA specification text, CC-BY-SA 4.0 clinical models, and an MPL-1.1-elected test corpus. **A gate that goes red on somebody else's licence teaches people to merge past a red gate** — which is exactly the failure this whole thread started with.

Results publish to the dashboard and are read there. Turning this into a gate is a decision to take deliberately once the six known findings are resolved, not a default inherited from an example.

## Token scope

The secret is a **push-only** token. `fossa analyze` documents that as sufficient (<https://docs.fossa.com/docs/cli/references/subcommands/analyze>: *"Works with a push-only API token"*), so it carries no read scope it does not need.

To answer the question that prompted this: the GitHub App installation does **not** authenticate the CLI. They are separate paths — the App scans server-side and ignores the config; the CLI runs on our runner and pushes up, so it needs its own key.

## Security posture

Every property the `zizmor` job enforces, verified rather than assumed:

- Both `uses:` pinned to full commit SHAs with the version in a trailing comment. Each SHA was dereferenced from its annotated release tag (`fossa-action` v2.0.0 → `29693cc`) and then **checked to belong to the repository named in front of it** — the `impostor-commit` class, which zizmor cannot verify in offline mode.
- `permissions: {}` at workflow level; the one job takes `contents: read`.
- `persist-credentials: false` on checkout.
- No context value reaches a `run:` block. `github.ref_name` is passed as an action *input*, not interpolated into a shell.
- No build cache restored — nothing here consumes build output.

`actionlint` (exit 0) and `zizmor --min-severity=low` (no findings) are both clean on the file.

## The Rust analyzer's requirements

FOSSA's Rust support runs `cargo generate-lockfile` then `cargo metadata --format-version 1` (<https://docs.fossa.com/docs/project-setup/supported-languages/rust>), so the job installs the pinned toolchain. The lockfile is committed, but the step runs defensively: a missing lockfile would otherwise surface as an empty dependency graph rather than an error.

Worth knowing when reading the results — from the same page, FOSSA *"does not currently understand default or missing features"* and treats optional dependencies like Java optionals. Its view of our graph is therefore **wider** than what actually compiles; `cargo deny` remains the precise instrument for what we ship.

## Requires

A `FOSSA_API_KEY` repository secret (push-only). Until it exists this job will fail on the analyze step — which is the honest failure mode, rather than silently reporting nothing.

Refs #2298.

### DIFF
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -1,0 +1,73 @@
+# Licence + dependency analysis via the FOSSA CLI (tracker issue #2298).
+#
+# WHY THIS LANE EXISTS, and not just the GitHub App: the hosted integration
+# scans server-side with its own credentials and does NOT read `.fossa.yml`
+# (that file is the CLI's — https://docs.fossa.com/docs/cli/references/files/fossa-yml).
+# Everything the committed configuration decides — that the dependency surface
+# is the Cargo graph, that the vendored openEHR specification text and third-
+# party test corpora are not ours to license, and the customLicenseSearch rules
+# that flag GPL-family text in FIRST-PARTY source — applies only when the CLI
+# runs. Without this lane the config is decoration and the dashboard reports
+# somebody else's licence as if it were ours.
+#
+# ANALYSE-ONLY, DELIBERATELY. `run-tests` stays false: `fossa test` would fail
+# the build on a policy verdict, and this repository redistributes vendored
+# material under CC-BY-SA and a tri-licensed test corpus. A gate that goes red
+# on somebody else's licence teaches people to merge past a red gate, which is
+# the failure this whole thread started with. Results are published and read on
+# the dashboard; flipping this to a gate is a deliberate decision to make once
+# the six known findings are resolved, not a default to inherit.
+#
+# The API key is a PUSH-ONLY token: `fossa analyze` "Works with a push-only API
+# token" (https://docs.fossa.com/docs/cli/references/subcommands/analyze), so
+# the secret carries no read scope it does not need.
+name: FOSSA
+
+on:
+  push:
+    branches: [develop, main]
+  pull_request:
+  workflow_dispatch:
+
+# Every job grants its own minimum; nothing is inherited.
+permissions: {}
+
+concurrency:
+  group: fossa-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  analyze:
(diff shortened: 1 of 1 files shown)


rubentalstra:
## Second commit: the pin was only half a pin

The action's README settles something the `action.yml` did not make obvious:

> This Action will always use the latest version of FOSSA CLI

So SHA-pinning `fossas/fossa-action` pinned the **wrapper** and left the tool that actually performs the analysis floating. Results could have changed with no commit in this repository — which is precisely what `.claude/rules/reliability.md`'s pinning requirement exists to prevent, and pinning the action alone only *looked* like compliance.

`pinned-cli-version: v3.17.16` now pins the CLI too, and it moves deliberately like every other pin here. Worth noting the README's own warning while reading it: *"Versioning of this action does not correspond to the version of FOSSA CLI"* — the two bump independently.

## Also added: an SPDX attribution report

`generate-report: spdx`, kept as a build artifact.

This is a different question from the SBOM the release lane already publishes. `release-build.yml` emits a CycloneDX SBOM per binary, which answers **what is in the artifact**. The attribution report answers **what we are obliged to attribute** — the question that actually matters for a project redistributing openEHR specification text, CKM clinical models and third-party test corpora under four different licences.

The report is a step *output*, so it exists only for the length of the job unless something writes it down. It is read through `env:` rather than interpolated into the shell: report content is arbitrary third-party text — dependency names and full licence bodies — and must never reach a shell as script.

`actionlint` exit 0, `zizmor --min-severity=low` no findings, and the `upload-artifact` SHA verified against its repository like the other two.

Issue #2312, opened 2026-08-12
ci(supply-chain): no SPDX SBOM reaches a consumer — it exists only as a workflow artifact


rubentalstra:
Once the release lane runs there will be a CycloneDX SBOM per release binary and an SPDX SBOM on each container image index. What will still not exist is an SBOM of the repository itself: the SPDX attribution report the licence-scanning lane already generates is uploaded as a workflow artifact and never attached to a release, never committed, and therefore reachable only by someone with access to a workflow run.

The two formats answer different questions for different readers, which is why publishing both is worth the small effort rather than being duplication. CycloneDX carries the dependency graph — direct versus transitive edges, package URLs, checksums — and is what a vulnerability tool consumes. SPDX is the attribution and licence-obligation document, and is what a legal or procurement reviewer of a four-licence redistribution actually asks for. For a clinical CDR both reviews happen, usually by different people, and neither should have to reconstruct the document from a CI run.

Two related notes for whoever picks this up. The Scorecard `SBOM` check is experimental and was **not** among the eighteen checks scored for this repository, so this work will not move the score and should not be justified on that basis — it is for the consumer. And the emitted CycloneDX specification version is worth revisiting against what current tooling consumes.

Effort: **S**.

References: [CISA — SBOM minimum elements](https://www.cisa.gov/sbom) · [SPDX specification](https://spdx.github.io/spdx-spec/) · [CycloneDX specification](https://cyclonedx.org/specification/overview/) · [Scorecard checks](https://github.com/ossf/scorecard/blob/main/docs/checks.md)

## Acceptance criteria

- [ ] The SPDX attribution document is attached to a release as an asset rather than existing only as a workflow artifact.
- [ ] The documentation states which SBOM answers which question, so a reader picks the right one instead of assuming they are redundant.
- [ ] The emitted CycloneDX specification version is a deliberate choice, recorded with its reason.
- [ ] Both documents describe the artifact actually shipped in that release, generated from the release commit rather than resolved afterwards.

rubentalstra:
**The issue's premise does not match the tree, and the tree is the authority.**

> the SPDX attribution report the licence-scanning lane already generates is
> uploaded as a workflow artifact and never attached to a release

No such report exists anywhere. `.github/workflows/fossa.yml` carries an explicit
block explaining why `generate-report` was **removed**: it needs a READ-scoped
token and fails with `Invalid API token type` on the push-only one this
[...]

Issue #2311, opened 2026-08-12
chore(legal): licensing does not survive file-level copying out of a four-licence tree (REUSE)


rubentalstra:
The repository redistributes material under four distinct licences and carries no `LICENSES/` directory, no `REUSE.toml`, and no per-file `SPDX-FileCopyrightText` / `SPDX-License-Identifier` headers. Licensing is currently expressed per vendored tree, by a provenance file naming the upstream source, the pinned revision and the licence, with the upstream licence text vendored alongside — plus an accurate licensing chapter, a scoped exception for one weak-copyleft build dependency, and a recorded election under one tri-licensed corpus.

That approach is genuinely good and should not be replaced; what it does not do is survive a file leaving the repository. The [REUSE Specification 3.3](https://reuse.software/spec-3.3/) exists for precisely that property — that licensing "is preserved when the file is copied and reused by third parties". Today, someone who lifts a single archetype out of a test corpus takes a CC-BY-SA-licensed file bearing no marking at all, and someone vendoring one generated crate source file takes a file whose dual-licence position is stated only in a manifest they did not copy. Given that the stated premise of this project is that people build on it, ship it and sell it, downstream redistribution of a four-licence tree is the expected case rather than the edge case.

The lever that makes this tractable is `REUSE.toml`, which expresses licensing in bulk by glob and therefore requires touching **no** vendored file — which matters, because hand-editing vendored material is forbidden by this project's own corpus discipline. The existing scan-exclusion list is already an accurate per-tree licence map and transposes almost mechanically. The real work is the generated crates: their headers must come from the emitter, never from a hand edit.

Effort: **M**.

References: [REUSE Specification 3.3](https://reuse.software/spec-3.3/) · [SPDX License List](https://spdx.org/licenses/) · [OpenSSF Best Practices criteria](https://www.bestpractices.dev/en/criteria/1) — `copyright_per_file`, `license_per_file`

## Acceptance criteria

- [ ] A `LICENSES/` directory holds the full text of every licence under which any file in the tree is offered, named by SPDX identifier.
- [ ] A `REUSE.toml` declares licensing for the vendored trees by glob, with no vendored file modified.
- [ ] First-party source carries SPDX copyright and licence headers, with the generated crates receiving theirs from the code generator rather than by hand.
- [ ] The elections and exceptions already recorded — the tri-licensed corpus, the weak-copyleft build dependency, the dual-licensed published crates — are represented rather than flattened.
[...]

Issue #2310, opened 2026-08-12
chore(security): secret scanning runs without non-provider patterns or validity checks


rubentalstra:
Secret scanning and push protection are enabled, but two sub-settings are off, verified live against the GitHub API: `secret_scanning_non_provider_patterns` is `disabled` and `secret_scanning_validity_checks` is `disabled`.

Provider patterns catch a GitHub token or a cloud access key. Non-provider patterns catch the credential classes this repository is actually most likely to leak: private keys, database connection strings with an embedded password, HTTP basic-auth URLs, and generic high-entropy secrets. That list is not hypothetical here — the Helm chart mounts a configuration volume that carries a private key, the OIDC configuration takes an HMAC secret with an enforced entropy floor, and the platform library ships a signing module. Those are precisely the values a hurried commit puts in a fixture, a compose override or a values file.

Validity checks report whether a detected credential is still live, which is the difference between "rotate this eventually" and "this is active right now". For a system holding PHI, that triage signal is worth having on by default.

Effort: **S** — two repository settings, no code.

References: [Supported secret scanning patterns](https://docs.github.com/en/code-security/secret-scanning/introduction/supported-secret-scanning-patterns) · [Secret scanning validity checks](https://docs.github.com/en/code-security/secret-scanning/managing-alerts-from-secret-scanning/evaluating-alerts) · [OpenSSF Security Baseline](https://baseline.openssf.org/versions/2025-02-25.html) — `OSPS-AC-01.01`

## Acceptance criteria

- [ ] Non-provider secret scanning patterns are enabled.
- [ ] Secret scanning validity checks are enabled.
- [ ] Any alert the newly enabled patterns raise against existing history is triaged, and a genuine finding is rotated rather than only dismissed.
- [ ] The enabled posture is recorded where the project's other repository-level security settings are described, so a later settings reset is detectable.

rubentalstra:
**Both settings are still `disabled`, and the REST API will not change them.**

Attempted three ways, all returning `200 OK` with the values unchanged — the
request is accepted and silently ignored:

```bash
# 1. both at once, JSON body
gh api -X PATCH repos/rubentalstra/FerroEHR --input - <<'JSON'
{"security_and_analysis":{"secret_scanning_non_provider_patterns":{"status":"enabled"},
                          "secret_scanning_validity_checks":{"status":"enabled"}}}
JSON

# 2 and 3. each on its own, field form
[...]

Issue #2309, opened 2026-08-12
chore(security): release tags carry no ruleset, so the publish lanes can be driven by an unprotected tag


rubentalstra:
Verified live against the GitHub API: the repository has two rulesets, `develop` and `protect-main`, both with `target: branch`. There is no ruleset with `target: tag`. Release tags are therefore unprotected against creation, movement and deletion, and three lanes trigger on a raw tag push — the release workflow, the chart publish workflow, and the documentation version-cut job.

Release immutability covers the window after a release is published, and it is correctly implemented here (draft, attach assets, verify the set is complete, then publish). But it does nothing before that point, which is exactly the window in which a tag drives a build, an image push and a chart publish. The chart lane already reasons about the layer above this one — it refuses to overwrite a published chart version precisely because an OCI tag is mutable — and the same reasoning applies to the git tag that triggers it. The tags are already GPG-signed and verify (`git tag -v v3.17.3` returns a good signature), so requiring signatures on the tag ruleset codifies a property the practice already has, and costs nothing.

This is the cheapest control in the programme relative to what it protects: the publish path for a clinical container image and a Helm chart on Artifact Hub.

Effort: **S**.

References: [Available rules for rulesets](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/available-rules-for-rulesets) · [About rulesets](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets)

## Acceptance criteria

- [x] A ruleset targeting release tags exists and is actively enforced — `release-tags`, `target: tag`, `enforcement: active`, `refs/tags/v*`.
- [x] Tag deletion and non-fast-forward tag updates are blocked — both rules present on the ruleset.
- [x] Tag signatures are required — `required_signatures`, and the v3.17.4 tag was signed and admitted.
- [x] The protection is verified on the LIVE path, not only by reading the settings back — pushing the signed `v3.17.4` tag exercised `required_signatures` and it was admitted.

[...]

PR #2317, opened 2026-08-12
chore(supply-chain): close the enterprise supply-chain and project-maturity gaps

### LINKED ISSUE #2305: documentation(security): no threat model, trust-boundary analysis or assurance case
There is no threat model, trust-boundary analysis or assurance case anywhere in the tree. Searching the README, the documentation book, the architecture reference and the security policy for `threat model`, `trust boundar`, `attack surface` and `STRIDE` returns two incidental prose mentions and no document. `OSPS-SA-03.01` (a security assessment identifying likely, impactful threats) and `OSPS-SA-03.02` (threat modeling and attack-surface analysis on critical paths) are unmet, as are the Best Practices silver `assurance_case` and `documentation_security` criteria.

This matters here for a specific, verifiable reason: the parts of this system that actually protect patient data are the parts no external specification governs. The architecture reference states plainly that openEHR ITS-REST makes authentication a `SHOULD` and mandates no scheme, and that the Service Model places authorization out of band — so only the 401-versus-403 split is spec-grounded. The five-stage authentication/`EHR_ACCESS`/RBAC/ABAC/SMART evaluation order, the fail-closed semantics, multi-tenancy isolation, the ATNA audit trail and the cold-archive tier are all this project's own design. A CDR whose authorization model is explicitly bespoke and carries no written threat model asks every deployer to redo that analysis privately, and asks every reviewer to reconstruct the intent from configuration prose.

Most of the raw material already exists, scattered across the security, identity-provider, audit and cluster-hardening chapters. The work is consolidation, explicit boundary naming, and — most valuably — stating the residual risk at each boundary rather than only the control.

Effort: **M–L**.

References: [OpenSSF Security Baseline](https://baseline.openssf.org/versions/2025-02-25.html) — `OSPS-SA-01.01`, `OSPS-SA-03.01`, `OSPS-SA-03.02` · [OpenSSF Best Practices criteria](https://www.bestpractices.dev/en/criteria/1) — `assurance_case`, `documentation_security`

## Acceptance criteria

- [ ] The actors and trust boundaries are enumerated: unauthenticated caller, authenticated clinical caller, admin/management caller, tenant peer, database, terminology server, object store, message broker.
- [ ] The protected assets are named explicitly, including clinical payload at rest, the audit trail, signing keys and bearer credentials.
- [ ] For each boundary, the control that holds is stated together with the residual risk that remains after it.
- [ ] The bespoke-by-necessity design decisions are identified as such, distinguishing what a specification requires from what this project chose.
- [ ] The document states what is explicitly NOT defended against, so silence is never read as coverage.
- [ ] The analysis is published on the documentation site, not held only in the repository.

### LINKED ISSUE #2306: documentation(security): no support window or end-of-life policy for released versions
The security policy states that the project is pre-release, that fixes land on `develop` and in the next tagged release, and that there are no maintenance branches yet. Meanwhile the project cuts official, non-prerelease GitHub releases — `v3.16.0` through `v3.17.3` inside eleven days — publishes a Helm chart carrying Artifact Hub Official status, and publishes eight crates to crates.io. Nothing anywhere states how long a released version receives security fixes, or when it stops receiving them. `OSPS-DO-04.01` (scope and duration of support per release) and `OSPS-DO-05.01` (when releases stop receiving security updates) are both unmet.

For a clinical deployment this is a planning question, not a formality. An operator running `v3.17.3` under change control needs to know whether a security fix will arrive as a patch on that line or only as the next minor with new behaviour attached — and with a release cadence measured in days and no LTS line, they currently cannot tell. The answer does not need to be generous to be useful: "only the latest release is supported, there are no maintenance branches, and a security fix ships as the next patch" is a complete and honest policy if that is what is true. What is not acceptable is the current combination of shipping official releases while describing the project as pre-release with an undefined support posture.

Effort: **S**.

References: [OpenSSF Security Baseline](https://baseline.openssf.org/versions/2025-02-25.html) — `OSPS-DO-04.01`, `OSPS-DO-05.01`

## Acceptance criteria

- [ ] The security policy states which versions receive security fixes, in terms that match the actual release practice.
- [ ] The policy states when a version stops receiving fixes and what an operator on it should do.
- [ ] The "pre-release, no maintenance branches yet" wording is replaced by the real policy or is made true.
- [ ] The same policy is reachable from the deployment documentation, not only from the security policy a researcher reads.
- [ ] The policy covers the Helm chart and the published crates, which follow their own version lines.

### LINKED ISSUE #2307: chore(security): accepted advisories are reasoned only in a comment, not published as VEX
Scorecard reports `Vulnerabilities: 4/10 — 6 existing vulnerabilities detected`. The advisory gate carries five reasoned, dated exceptions (RUSTSEC-2023-0071, RUSTSEC-2026-0194, RUSTSEC-2026-0195, RUSTSEC-2026-0173, RUSTSEC-2024-0436), each with the exposure argument and a re-check condition, plus one deliberately un-ignored advisory that the feature-resolving tool never raises. The reasoning is good; the problem is that it lives only in a TOML comment, so an SCA tool scanning a deployed artifact sees six unexplained findings and the argument that was already made reaches nobody downstream. `OSPS-VM-04.02` asks for exactly this: non-affecting vulnerabilities accounted for in VEX documents.

This is worth doing here because the machinery already exists and is already used for the harder case — the repository publishes an OpenVEX document for an inherited container-layer finding, with the justification and an impact statement a reader can check. Extending that to the Rust advisories converts five paragraphs nobody downstream reads into a document a hospital's scanner ingests. For a clinical deployer, "six unexplained CVEs" and "six CVEs with a published not-affected justification" are entirely different procurement conversations, and the second one is already true — it is just not published.

Generating the documents from the advisory-gate configuration rather than hand-writing them is the point: two hand-maintained lists drift, and this repository has already been bitten once by a second advisory ignore list that nothing read.

Effort: **S**.

References: [OpenSSF Security Baseline](https://baseline.openssf.org/versions/2025-02-25.html) — `OSPS-VM-04.02` · [OpenVEX specification](https://github.com/openvex/spec) · [Scorecard — Vulnerabilities](https://github.com/ossf/scorecard/blob/main/docs/checks.md#vulnerabilities)

## Acceptance criteria

- [ ] Each accepted advisory has an OpenVEX statement with a status and a justification drawn from the specification's vocabulary.
- [ ] The justification matches the actual argument: code not present in the built artifact where features exclude it, and adversary-uncontrollable where the operation is verification-only.
- [ ] The documents are generated from the advisory-gate configuration so the two cannot disagree.
- [ ] A drift check fails if an accepted advisory has no corresponding statement.
- [ ] The documents are attached to a release so a deployer's scanner can consume them.

### LINKED ISSUE #2308: ci(deps): Dependabot covers neither the fuzz workspace nor the digest-pinned base images
The Dependabot configuration declares exactly two ecosystems, `cargo` at `/` and `github-actions` at `/`, and two dependency surfaces fall outside both.

The fuzz harnesses are their own Cargo workspace with their own lock file, deliberately excluded from the root workspace members. A `cargo` entry at `/` does not reach a separate workspace, so those dependencies never receive an update proposal and never receive a security bump — while the harnesses are compiled on the pull-request path specifically so a scheduled-only lane cannot rot unnoticed.

The three container base images are pinned by digest — the PostgreSQL image, the Rust builder and the distroless runtime. Digest pinning is correct for reproducibility and is what earns the Scorecard `Pinned-Dependencies: 10`, but it has a cost that only an update lane pays off: a digest-pinned base is a base that never receives a patch unless a human remembers. With no `docker` ecosystem entry, the weekly published-image scan will keep *finding* the resulting CVEs in the clinical images while nothing proposes the fix. That is the gap between detection and remediation, and it sits on the images an operator actually runs.

Effort: **S**.

References: [Dependabot options reference](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference) · [Scorecard — Dependency-Update-Tool](https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool)

## Acceptance criteria

- [ ] The fuzz workspace receives Dependabot cargo updates against its own lock file.
- [ ] Each Dockerfile directory receives Dependabot docker updates, so pinned base-image digests are proposed for bump rather than aging silently.
- [ ] New entries carry a cooldown consistent with the existing ones, and the reasoning for the chosen window is recorded where the existing windows are explained.
- [ ] A base-image digest bump proposed by the bot passes the existing image build and scan lanes without manual intervention.

### LINKED ISSUE #2313: documentation: two committed claims have drifted from the tree they describe
Two committed statements no longer match the tree, both found while auditing the supply-chain posture. Neither is a security defect; both are the kind of drift that makes a reader distrust the surrounding text, which in a repository whose credibility rests on machine-verified claims is worth more than the size of the fix suggests.

First, the reliability rule set justifies keeping the API-shape lints active on the grounds that every crate is unpublished and therefore has no external semver surface to protect. That was true when written and is now false for the eight specification crates, which publish to crates.io on their own version line — so the sentence argues from a premise that no longer holds, in the same document that adjudicates how public API stability is treated.

Second, the security chapter documents release verification against a version that does not yet exist. The chapter is honest about this and flags in-page that those forms begin answering at the next cut, so it is not a false claim — but it becomes stale in the other direction the moment that release lands, and the examples should then be confirmed against real assets rather than assumed correct.

Effort: **S**.

## Acceptance criteria

- [ ] The unpublished-crates justification is corrected to reflect that the specification crates are published, without changing the lint decision it supports unless that decision genuinely changes.
- [ ] The release-verification examples name a release that exists and have been executed against its real assets.
- [ ] The in-page caveat about forms that do not answer yet is removed once they do, so the caveat does not itself become the stale statement.

## Tasks

- [ ] Re-run each documented verification command against the cut release and paste nothing that was not observed.

rubentalstra:
Closes #2305
Closes #2306
Closes #2307
Closes #2308
Closes #2313

Eleven sub-issues of #2302. Five close here. Four are partially delivered and say
so below; two need a repository setting only the owner can flip, with the exact
steps posted on their issues. Nothing is claimed that was not observed.

### #2305 — threat model

A new `website/book/src/threat-model.md` chapter, linked from the security
chapter and the README. It names the actors, the assets in the order the loss
hurts, and eight trust boundaries; for each boundary it states the control **and
the residual risk that survives it**, which is the part the existing
configuration prose could not give a reader. It ends with an explicit
not-defended-against list, so silence is never read as coverage, and with the
rule that where the document and the code disagree the code is right and the
document is a defect.

Grounded first-hand rather than from the issue text: the five-stage evaluation
order and its fail-closed semantics from `docs/architecture.md`; tenant isolation
read from `app/ferroehr/migrations/ehr/0004_multitenancy.sql` (`FORCE ROW LEVEL
SECURITY` driven by the `ferroehr.tenant_id` session setting) and
`app/ferroehr/src/extensions/tenancy.rs`; the audit chain from the audit chapter;
the supply-chain boundary from the release lanes in this same PR.

### #2306 — support window and end-of-life

`SECURITY.md` § Supported versions replaced. Only the newest release is
supported; there are no maintenance branches and no backports; a version stops
receiving fixes the moment a newer release exists; a fix normally arrives as the
next patch but that is not promised. Covers the server, the Helm chart and the
crates on their separate version lines, and closes with what to do if that is not
enough. The same policy is repeated in the Operations chapter's upgrade guidance,
where an operator planning change control actually looks.

### #2307 — VEX for the accepted advisories

`security/vex/rust-advisories.openvex.json`, **generated** by
`scripts/security/vex-generate.sh` from `deny.toml` (the id set — the gate that
actually decides whether a build passes) joined with
`security/vex/rust-advisories.toml` (the prose). The generator refuses to emit
anything if the two disagree in either direction, and `vex-advisories.sh` — the
new `vex` CI job — regenerates and diffs, so a hand edit fails too.

Justifications match the real arguments and come from the OpenVEX vocabulary:
`vulnerable_code_cannot_be_controlled_by_adversary` for `rsa` (verification-only,
no private keys held) and the two `quick-xml` advisories (parses
operator-configured S3 responses, never client input);
`vulnerable_code_not_present` for the two proc-macro advisories, which link into
nothing shipped. It also carries `RUSTSEC-2026-0235` (`rkyv`,
`component_not_present`), which `cargo-deny` never raises and a
`Cargo.lock`-reading scanner does — the class of finding that reaches a consumer
with no explanation anywhere in the repository, and the reason publishing this is
worth anything.

Mutation-proven, five ways, each failing and the restored tree passing: an
accepted advisory with no prose entry; a prose entry the gate no longer ignores;
one hand-edited byte in the JSON; a lock-file-only entry added to the gate; an
out-of-vocabulary justification.

### #2308 — Dependabot coverage

A second `cargo` entry at `/fuzz` (separate workspace, separate lock file, which
a `/` entry cannot reach) and a `docker` entry over the three Dockerfile
directories. Verified against the current options reference: separate entries per
ecosystem are supported while directories do not overlap; `directories` (plural)
takes a list where `directory` does not; and `cooldown` applies to version
updates only, never to security updates — which is what makes the windows safe.

### #2313 — drifted claims

`clippy.toml`, `.claude/rules/reliability.md` and `deny.toml` all justified a
decision with "every crate is `publish = false`". False since the eight
`openehr-*` crates started publishing. The lint decision is unchanged because the
real reason is better: they publish on `0.0.x`, where cargo treats every patch as
its own compatibility set, so no consumer range spans a break. The `deny.toml`
variant was doubly wrong — cargo refuses to package a path dependency without a
version, which is why the published crates carry explicit lockstep versions and
do not rely on `allow-wildcard-paths` at all.

The security chapter's verification examples named `v3.17.4`, which does not
exist. They now write `<tag>`, and the in-page caveat states the real position:
no published release carries these assets yet, the newest is `v3.17.3` from
2026-08-05, and the commands are to be re-run against real assets at the next
cut.

### #2303 — signed releases (the lane still has to run)

Implemented now:

- **`.sha256sum` per tarball.** The only verification available with neither `gh`
  nor `cosign` installed. Documented as detecting a corrupt download and **not** a
  substituted release. Checked against Scorecard's actual probe source before
  claiming anything: `probes/releasesAreSigned` matches `.asc`, `.minisig`,
  `.sig`, `.sign`, `.sto`, `.sigstore.json` and `probes/releasesHaveProvenance`
  matches `.intoto.jsonl` — no checksum suffix, so the checksum earns no score and
  is not sold as if it did. The existing `.sigstore.json` and `.intoto.jsonl`
  assets are what will move `Signed-Releases`.
- **The `finalize-release` guard, read carefully as asked — and it had a real
  defect.** The names do agree, but the comparison was `grep -q` on a
  **substring**, so the check for `.tar.gz` was satisfied by
  `.tar.gz.sigstore.json` alone: a draft missing its actual tarball would have
  been published, unfixably. Now `grep -Fxq`, whole-line, plus the checksum, the
  repository SBOM and the compose files.

Not closable here: the lane has never executed, and only a real cut can satisfy
"confirmed to have run and passed", "verified against a release that exists" and
"a subsequent Scorecard run reports non-zero". Those three acceptance boxes, and
#2313's task to re-run each documented command against real assets, belong to the
next release cut.

### #2311 — REUSE

`LICENSES/` (six texts, SPDX-named) and a root `REUSE.toml`. `reuse lint` reports
the project **compliant with version 3.3** — 13551/13551 files with copyright and
licence information — and it gates the merge via the official image, digest-pinned
like the actionlint one. **No vendored file was modified**, which is why the glob
form was the only available mechanism.

Both recorded positions are represented rather than flattened: the three
tri-licensed ISO 13606 BMM models are declared `MPL-1.1` per the existing
election, and `PropertyUnitData.xsd` — an AGPL header inside a CC-BY-SA-3.0
repository — is declared at the more restrictive of the two readings, the one the
file's own text asserts.

A second gate, `licensing-declarations.sh`, fails the build when `REUSE.toml`,
`LICENSES/` and the licensing chapter stop agreeing about which licences apply
(three mutations proven). The copyleft gate needed one adjustment and got it
without weakening: `LICENSES/**` joins the PROSE list beside the root licence
texts it is the same category as, and a licence text belonging to no file now
fails the new gate instead — re-proven that a GPL grant dropped into
`app/ferroehr/src` still fails.

**Left open:** SPDX headers in first-party source. Sub-issue #2318 carries it,
because it is a ~2500-file mechanical change that must go through the
`openehr-codegen` emitter for the 1471 generated files, forces a full
regeneration and a lockstep bump of all eight published crates, and does not
belong in a pull request that also lands governance documents and release-lane
changes. Note that for the case the issue actually motivates — an archetype
lifted out of a corpus — headers were never available: vendored files may not be
edited, which is exactly why `REUSE.toml` is the answer there.

### #2312 — SBOM to a consumer

A release now carries `ferroehr-<tag>.spdx.json`, generated by syft (pinned
1.50.0) from the checked-out release commit, attached to the draft and required by
the completeness guard. The security chapter gains a table saying which of the
three SBOMs answers which question, so a reader picks one instead of assuming they
are redundant. The CycloneDX version is recorded with its reason: 1.5 is the
**highest** `cargo-cyclonedx` emits (it accepts 1.3/1.4/1.5 and defaults to 1.3),
so the flag is load-bearing and 1.6 is simply not reachable from this generator.

**The issue's premise is wrong and the tree is the authority.** It says "the SPDX
attribution report the licence-scanning lane already generates is uploaded as a
workflow artifact". No such report exists: `generate-report` was deliberately
removed from `fossa.yml` (measured 2026-08-12) because it needs a read-scoped
token this repository does not issue, and the reasoning is in that file. So there
was no SPDX SBOM anywhere, not one stuck in a workflow artifact — hence a
generator rather than a re-upload. The remaining criterion needs a cut, like
#2303.

### #2304 — governance

`GOVERNANCE.md`, `MAINTAINERS.md`, `SUPPORT.md`, `.github/CODEOWNERS`, and a
README section. They state the single-maintainer reality plainly: bus factor one,
no organisation, no legal entity, and a publishing-identity table whose recovery
column reads "none" for most rows because that is true. `GOVERNANCE.md` says
outright that the maintainer's own changes get no independent human review and
that machine enforcement stands in for it, rather than describing a review process
that does not happen.

**Left open — needs the owner:** the two ruleset criteria. The `PUT` on the
`develop` ruleset was refused by this session's permission layer. The exact call
is on the issue.

## What needs the owner

- **#2309 — done in part, verification outstanding.** The `release-tags` ruleset
  (id `20755150`) was **created and is active**: `refs/tags/v*`, blocking deletion
  and non-fast-forward updates and requiring signatures. The criterion "verified by
  attempting a disallowed tag operation" is **not** met — the only safe probe
  needed a ruleset edit that was refused, and no destructive probe was run against
  a real release tag. Steps on the issue.
- **#2310 — blocked on a UI toggle.** `secret_scanning_non_provider_patterns` and
  `secret_scanning_validity_checks` are still `disabled`. `PATCH
  /repos/{owner}/{repo}` **accepts and silently ignores** both, together and
  individually, returning `200` with the values unchanged — verified three times.
  They have to be switched on in Settings. `SECURITY.md` now carries the full
  expected repository security posture as a table with the read-back commands, and
  marks these two as not-yet-true rather than describing them as if they were.

## Verification

Run on this branch, results as observed:

- `cargo fmt --all` — clean; zero `.rs` files changed by this PR.
- `cargo clippy --workspace --exclude ferroehr-admin-ui --all-targets --all-features -- -D warnings` — exit 0.
- `bash scripts/checks/ci-conclusion-complete.sh` — `ok: all 39 CI jobs gate the merge` (the two new jobs, `vex` and `licensing`, are in `conclusion.needs`).
- `bash scripts/checks/no-python.sh` — clean. `reuse` reaches CI as the official container, never a Python install.
- `SHELLCHECK_OPTS='-e SC2016' actionlint` — no output.
- `zizmor --min-severity=low .github/workflows/` — `No findings to report`.
- `reuse lint` — `compliant with version 3.3`, 13551/13551.
- `vex-advisories.sh`, `licensing-declarations.sh`, `first-party-license-text.sh`, `docs-claims.sh --all`, `comment-style.sh --all` (2462 files), `changelog-structure.sh` — all pass.
- Every new gate mutation-proven in both directions; the mutations are listed per issue above.
- All commits GPG-verified.

## Findings for the tracker

1. **A tracked filename contains a literal newline.**
   `docs/specs/openehr/CNF/tests/platform/robot/I_EHR_COMPOSITION/get_composition/I_EHR_COMPOSITION.get_composition_version-bad_ehr\n.robot`
   — confirmed by dumping `git ls-files -z`. Glob matching does not reach it, so
   `REUSE.toml` has to name it explicitly. Upstream's defect, arriving through the
   vendoring script; a candidate `upstream-report`.
2. **`target/` is 71 GB**, well past the ~30 GB `cargo clean` threshold. Not
   cleaned: another session shares this tree and the cache is warm.
3. **The ADL2 library's licence claim is unverified.** Two PROVENANCE files say
   "predominantly CC-BY-SA 3.0 where stated" for corpora whose archetypes write
   `Creative Commons CC-BY-SA` with the version only in a URL. The CKM claims were
   measurably wrong and are fixed in this PR; this one was left alone rather than
   guessed at.

### DIFF
--- a/.claude/agent-memory/cnf-triage/MEMORY.md
+++ b/.claude/agent-memory/cnf-triage/MEMORY.md
@@ -41,3 +41,6 @@
 - [OAS `required:` ≠ non-empty array](oas-required-does-not-forbid-empty-arrays.md) — `Clstr.yaml required: [items]` is satisfied by `[]`; only the BMM cardinality grounds an empty-list refusal
 - [Authz class is EXPLICIT SM deferral](authz-class-is-explicit-sm-deferral.md) — a role 403 is never the app bin; SM master02 §Functional Style also grounds driving semantics cases as an authorized principal
 - [Definition-delete routes are Admin (#2071)](definition-delete-routes-are-admin-class.md) — the class move stands; 5 red rows = catalogue bin (`on: admin`), plus the readonly/uncased-branch fallout
+- [TERMINOLOGY_ID has no declared invariant](terminology-id-value-has-no-invariant.md) — BASE declares no Invariants row; master05's production is refuted by its own examples AND released QUERY `snomed_ct(3.1)`
+- [PGP subkey signature the runner cannot verify](pgp-subkey-signature-runner-cannot-verify.md) — RUNNER bin: rpgp 0.20 `SignedPublicKey::verify` is primary-only; the app signs with the `[S]` subkey since 7a8a8c9a3
+- [Baseline commit = where the run was RECORDED](baseline-commit-is-where-the-run-was-recorded.md) — diff the whole range since the previous results.json, never trust a handed-down "prime suspect"
--- a/.claude/agent-memory/cnf-triage/baseline-commit-is-where-the-run-was-recorded.md
+++ b/.claude/agent-memory/cnf-triage/baseline-commit-is-where-the-run-was-recorded.md
@@ -1,0 +1,24 @@
+---
+name: baseline-commit-is-where-the-run-was-recorded
+description: A results.json regression is attributed to the commit that RE-RAN the pipeline, not the commit that caused it — always diff the range since the previous baseline
+metadata:
+  type: feedback
+---
+
+`docs/conformance/ferroehr/results.json` is regenerated only when someone runs
+`scripts/conformance.sh`, not per PR. So "the regression landed in commit X" is
+almost always false: X is where the run was RECORDED.
+
+**Why:** on 2026-08-12 the 8 red rows were handed over as "landed in 11ee41ea7
+(#2293), the prime suspect". `git log --oneline 6ffd14ed6..11ee41ea7 --
+docs/conformance/ferroehr/results.json` shows 11ee41ea7 is the ONLY commit in
+the range that touched it — and the range is **145 commits**. Neither cluster's
+cause was in #2293: the terminology cluster came from bdf68d3f1 (#2282) and the
+PGP cluster from 7a8a8c9a3, both earlier in the range.
+
+**How to apply:** step 0 of any regression triage is
+`git log --oneline <prev-baseline>..<red-baseline> -- docs/conformance/<sut>/results.json`
+to establish the true suspect RANGE, then
+`git diff <prev>..<red> -- <the module the failure implicates>` to find the one
+semantic change. Never accept a handed-down "prime suspect" commit without
+that check.
--- a/.claude/agent-memory/cnf-triage/pgp-subkey-signature-runner-cannot-verify.md
+++ b/.claude/agent-memory/cnf-triage/pgp-subkey-signature-runner-cannot-verify.md
@@ -1,0 +1,40 @@
+---
+name: pgp-subkey-signature-runner-cannot-verify
+description: RUNNER bin — rpgp 0.20 SignedPublicKey::verify is primary-key-only, so the runner rejects the app's (conformant) signing-subkey signature
+metadata:
+  type: project
+---
+
+Confirmed 2026-08-12: all three `SIG-VERSION-*-pgp` cases asserting
+`verifiable: true` fail with "signature: does not verify over the agreed
+canonical form"; every digest twin and every mode-agnostic pgp case passes.
+
+The chain, all read first-hand:
+
+1. `app/ferroehr/src/versioning/signature/key.rs:157-160` selects a
+   signing-capable SUBKEY by RFC 9580 §5.2.3.29 key flag 0x02 and signs with it
+   (`:181-195`). Landed in **7a8a8c9a3** (2026-08-07) — AFTER the last green
+   baseline; before it the primary key signed.
+2. The committed test certificate HAS one: `gpg --show-keys
+   tools/cnf-runner/artifacts/corpus/keys/cnf-signing.sec.asc` →
+   `sec [SCEAR] C87EEF94…` + `ssb [S] 376C44DE…`. The ixit's
+   `instances.sut_pgp.signing.public_key` is byte-identical to the committed
+   `.pub.asc`, i.e. the same certificate.
+3. **Pinned `pgp` 0.20.0, `src/composed/signed_key/public.rs:200-204`:**
+   `impl VerifyingKey for SignedPublicKey { fn verify(..) { self.primary_key.verify(..) } }`
+   — `public_subkeys` is never consulted.
+4. `tools/cnf-runner/src/exec/signature.rs:126` is
+   `sig.verify(&key, bytes).is_ok()` on the primary `SignedPublicKey` → `Ok(false)`.
+
+The app already has the correct shape:
+`key.rs::verify_against_certificate` (`:224-236`) tries the primary key then
+every subkey. The runner needs the same walk.
+
+Not environmental: `PgpKey::load` is fail-closed at boot (a test signature), so
+the pgp deployment could not have started with a missing/unusable key, and
+`SIG-VERSION-client_supplied_verbatim-pgp` passed on the same instance in the
+same run. `Ok(false)` (not `Err`) also proves the armor PARSED.
+
+**How to apply:** a `-pgp` verifiable failure with the digest twin green is
+never a canonicalization bug — the digest case proves the JCS bytes agree. Look
+at which key component signed.
--- a/.claude/agent-memory/cnf-triage/terminology-id-value-has-no-invariant.md
+++ b/.claude/agent-memory/cnf-triage/terminology-id-value-has-no-invariant.md
@@ -1,0 +1,52 @@
+---
+name: terminology-id-value-has-no-invariant
+description: BASE declares NO Invariants row on TERMINOLOGY_ID; the master05 §Syntaxes production is self-refuted by its own examples and by released QUERY `snomed_ct(3.1)`
+metadata:
+  type: project
+---
+
+Confirmed first-hand 2026-08-12 while triaging the 5-row terminology cluster
+(build 11ee41ea7; the enforcement landed in bdf68d3f1 / issue #2258).
+
+**The released model does not declare the invariant the app enforces.**
+`BASE/docs/UML/classes/org.openehr.base.base_types.terminology_id.adoc` has NO
+`*Invariants*` row at all — its only value statement is the description line
+"Lexical form: `name [ '(' version ')' ]`". The sibling
+`…version_tree_id.adoc`, whose lexical form is described the same way, DOES
+carry an Invariants row — and even there `Value_valid` is only
+`not value.is_empty`. The BMM mirrors this exactly
+(`openehr_base_1.3.0.bmm.json`: `VERSION_TREE_ID.invariants` = 7 entries,
+`TERMINOLOGY_ID` has no `invariants` key; same for ARCHETYPE_ID, TEMPLATE_ID,
+GENERIC_ID, OBJECT_VERSION_ID, HIER_OBJECT_ID). Neither released ITS
+constrains the value beyond `string` (ITS-JSON
+`openehr_rm_1.1.0_all.json` `TERMINOLOGY_ID.properties.value: {type: string}`;
+ITS-XML `BaseTypes.xsd` `TERMINOLOGY_ID` = bare `xs:extension base="OBJECT_ID"`).
+
+**The production is refuted by released text, twice.**
+`master05-identification_package.adoc:273/277`
+`terminology_id = name-str, [ '(', name-str, ')' ]`,
+`name-str = letter, { letter | digit | '_' | '-' | '/' | '+' }`.
+(a) Its own §Terminology Identifiers examples `ICD9(1999)`, `ICD10AM(3rd_ed)`,
+`ICD10AM(4th_ed)` all start the version with a digit (upstream report #2283).
+(b) **RELEASED QUERY 1.1.0** `docs/AQL/master03-syntax.adoc:239` publishes
+`name/defining_code/terminology_id/value='snomed_ct(3.1)'` as the canonical
+expansion of the node-predicate shortcut — `.` is outside `name-str`, so #2283's
+"one rule wide" relaxation is one rule too narrow. #2283 does not mention this.
+
+**Split reading that survives the released text:** the NAME part's production is
+NOT contradicted by any released example (`SNOMED-CT`, `ICD10`, `ICD9`,
+`ICD10AM` all conform), so refusing an interior space is defensible; the VERSION
+part's character class is contradicted; and a URI-form id
+(`http://…/CodeSystem/x`) is released-SILENT — grep of RM/BASE/QUERY/TERM/
+ITS-REST finds no URI spelled as a `terminology_id` value (the `http://snomed.info/sct`
+occurrences in QUERY §TERMINOLOGY are function arguments and `code_string`
+operands, never a TERMINOLOGY_ID).
+
+**How to apply:** any red row reading "expected `created`, observed
+`validation_failed`" on a fixture whose CODE_PHRASE names an external
+terminology — check the TERMINOLOGY_ID value against
+`crates/openehr-base/src/v1_3/base_types/identification/terminology_id_impl.rs`
+`is_valid_terminology_id` first. Fix path is the TEMPLATE
+`tools/openehr-codegen/templates/openehr-base/base_types/identification/terminology_id_impl.rs`
++ `openehr-codegen -- emit` (both stamped copies are
+`@generated-from-template`). See [[cnf-red-2026-08-12-terminology-id-grammar]].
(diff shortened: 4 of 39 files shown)


rubentalstra:
**The red `ui-e2e` check is #2285, not this PR.**

`e2e_admin_ops::admin_deletes_a_template_from_the_detail_screen` fails with
`` `minimal_instruction.en.v1` never appeared after uploading
`minimal_instruction.opt` (pre-hydration uploads exhausted) `` — byte-identical to
the failure on **`develop`** in run `31614565671`, same test, same panic, same
line. It reproduces on `fix/cnf-subkey-verification` and `ci/green-lanes` too.

This branch touches no console source, no server source, no compose file and no
Dockerfile — zero `.rs` files change in the whole PR. The job ran here only
because the `admin_ui` change filter deliberately includes
`.github/workflows/ci.yml`, which this PR edits to add the `vex` and `licensing`
jobs; on `develop` the filter usually skips it, which is why a red gate has been
sitting there quietly.

Nothing was skipped, weakened or retried to get around it. It clears when #2285
does.

Every other job passed, including both new gates (`vex`, `licensing`).

Issue #2780, opened 2026-08-26
ci(workflows): phase 6 — the verification cut through the new pipeline


rubentalstra:
Phase 6 of the greenfield estate redesign (design: `docs/plans/github-workflows-greenfield.md` §4.6) — the proof. A `vX.Y.Z-rc1` tag runs the whole new pipeline end to end: plan green, binaries + images + scan + apply-tags, chart, sandbox, docs-freeze, announce — asserting the PRERELEASE semantics everywhere (`:latest` does not move, `/docs/latest/` does not move, the GitHub Release is marked prerelease, the sandbox does not deploy). Then the first REAL cut on the new pipeline is the closing evidence for the whole program.

## Acceptance criteria
- [ ] The rc tag's pipeline run is green end to end with prerelease semantics proven at every leg
- [ ] The first real release tag after it is green end to end with zero manual interventions and zero red lanes anywhere in the estate
- [ ] The design plan file is deleted in the PR that closes this phase (the delete-on-implementation rule); the program's durable record is #2772 + the phase PRs

Issue #2772, opened 2026-08-26
ci(workflows): greenfield redesign of the entire workflow estate — one designed system instead of 23 organic files


rubentalstra:
The 23 workflows under `.github/workflows/` grew organically, bolted on one need at a time, and the seams show: the chart lane polls the Containers pipeline in-job and died on a transient TLS handshake timeout on two consecutive release tags (#2771); the sandbox path needed its own redesign after racing the image lane (#2724, #2751, #2755); a release cut hand-edits ~10 pin sites each guarded by a differently-timed check; boilerplate, secrets wiring, and shared constants are copy-pasted across files. Owner directive 2026-08-26: design the whole estate greenfield — how everything flows, every workflow rethought, nothing untouched — BEFORE any changes land.

The deliverable of this issue is the DESIGN: a full inventory-grounded plan (`docs/plans/github-workflows-greenfield.md`) covering the complete event flow (PR / develop push / release tag / schedule / manual), the orchestration primitives to standardize on (reusable workflows and `needs` graphs over cross-workflow polling; `workflow_run` only where a real boundary exists), retry discipline so a transient network failure can never red a lane, the publish/deploy ordering (build → scan → tag → chart → sandbox → release, each gated on the previous by construction), a single source of truth for versions and shared constants, the security posture applied uniformly, and the failure/recovery story for every lane. Grounded in the current estate's measured defects and in industry patterns with sources named.

Implementation happens as sub-issues per phase AFTER the owner approves the design; the plan file is deleted by the PR that completes the last phase, per the delete-on-implementation rule.

## Acceptance criteria
- [x] The design plan exists, covers all 23 workflows plus the cut procedure and its guards, and is grounded in the first-hand inventory (every trigger, race, poll, duplication, and manual recovery named)
- [x] Every cross-workflow ordering in the design is event-driven or `needs`-graphed — no in-job polling of other workflows anywhere
- [x] The design names its industry-pattern sources for each structural choice
- [x] Owner has reviewed and approved the design
- [x] Implementation sub-issues are filed per phase, each independently shippable

rubentalstra:
[...]

Issue #2011, opened 2026-08-06
ci(security): no cache in the release path (artifact poisoning)


rubentalstra:
**Control: "Disable all forms of caching in release or publishing workflows to avoid reusing potentially compromised artifacts"** — the artifact-poisoning class.

**Finding: `release.yml` uses caching.** A release lane that restores a cache can build a published artifact from bytes produced by an earlier, less-trusted run (for example a pull-request build), which is exactly the reuse the cheat sheet warns about. `publish-crates.yml` appears clean and should be confirmed as such.

This needs judgement rather than a blanket removal: the release build is long, and the cache exists for a reason. The honest options are to drop caching from the release path outright, or to scope the cache so a release can only ever restore entries produced by a release build on a protected ref — and to say which was chosen and why.

## Acceptance criteria
- [x] The release path either uses no cache, or uses one provably unable to restore entries from an untrusted run
- [x] `publish-crates.yml` confirmed cache-free
- [x] The reasoning recorded, including the build-time cost accepted

PR #2790, opened 2026-08-26
ci(release): one needs-ordered pipeline owns the v* tag

### LINKED ISSUE #2771: ci(chart): redesign the chart-publish lane — event-driven on the image publish, no in-job polling
The chart-publish lane has gone red on two consecutive release tags and needs the same redesign the sandbox delivery got in #2724: event-driven ordering instead of a poll loop.

What the runs show, first-hand:
- v4.0.4 (run 32935065518): the lane triggers on the tag push, then busy-waits inside the job for the tag's Containers "app image" job so it can verify the appVersion image against the chart's rendered defaults. One transient `TLS handshake timeout` on a poll call to the GitHub API killed the run under `set -e` — a pure infra flake turned into a red release lane.
- v4.0.3 (run 32867218726): same lane, also failure; the only successful publish in the sequence was a manual `workflow_dispatch` after the fact.

The structural problems, independent of the flake:
- A tag-triggered job that waits ~40 minutes for a sibling workflow burns a runner the whole time and races the thing it waits for — the same shape #2723/#2724 removed from the sandbox path.
- Unretried GitHub API calls inside the wait make every network hiccup fatal.
- A red chart lane on a tag looks like a failed release even when the chart content is fine, and the recovery is a hand dispatch nobody is told about.

Redesign direction (decide at pickup): publish the chart from a `workflow_run` trigger on the Containers pipeline's success for release tags — after `apply-tags`, the appVersion image is guaranteed pullable, so the existence check becomes a plain pull instead of a wait — or fold the publish in as a job of containers.yml behind `apply-tags`. Either way: no in-job polling of other workflows, retries on the remaining API/network calls, and the existing safety properties preserved exactly (never overwrite a published chart version; queue-never-cancel concurrency; Sigstore attestation; the Artifact Hub ownership tag; the dry-run dispatch lane).

## Acceptance criteria
- [ ] No job in the chart lane waits on another workflow by polling
- [ ] The appVersion-image check runs only when the image is already published, by construction of the trigger
- [ ] A transient network failure retries instead of failing the lane
- [ ] The v4.0.4/v4.0.3-class failure is reproduced as impossible: a release tag's chart publish is green end-to-end on the next cut, with no manual dispatch
- [ ] The safety properties are re-verified: refuse-overwrite, queue-never-cancel, attestation, ownership tag, dry-run dispatch

### LINKED ISSUE #2776: ci(workflows): phase 2 — the needs-ordered release pipeline
Phase 2 of the greenfield estate redesign (design: `docs/plans/github-workflows-greenfield.md` §2.2) — the core: ONE workflow owns the `v*` tag, every ordering a `needs` edge, no workflow anywhere polls another.

Shape: `plan` (re-runs the ENTIRE release-fact guard tier against the tagged commit before anything builds — closes the tag-time guard hole and the pending-cancellation blind spot) → binaries (release-build.yml ×2 arch) and images (build-image.yml ×3, defer-tags) → image scan → apply-tags (moves :latest) → in parallel: chart publish (build-chart.yml, `needs: apply-tags` — the appVersion image is pullable by construction, the 60-minute poll loop and its job-name coupling are DELETED), sandbox deploy+reseed (sandbox-deploy.yml gains `workflow_call:`, keyed on apply-tags success, never on a run conclusion — an advisory scanner can never again suppress a deploy), and docs-freeze (moved out of docs.yml; prerelease-guarded; IDEMPOTENT — an already-cut version is a green no-op) → `announce` (aggregate: asserts every leg, and on failure names the exact recovery). Every leg re-runnable for the tag; the chart leg keeps its full safety chain verbatim (refuse-overwrite, cosign retry, read-back, ownership tag).

Removals in the same phase: the tag triggers on containers.yml, publish-chart.yml (the caller file goes entirely), docs.yml's version-cut job, and sandbox-deploy.yml's workflow_run trigger. The SLSA-L3 construction, attestations, draft-then-publish immutability shape, and scan-before-tags are preserved exactly.

## Acceptance criteria
- [ ] One workflow triggers on `v*`; `gh run list` for a tag shows exactly one release-pipeline run plus zero tag-triggered legacy runs
- [ ] No job in the estate polls another workflow's runs or jobs via the API (grep for the idiom finds nothing)
- [ ] `plan` fails a deliberately broken pin (mutation test) before any build starts, and a green `plan` is followed by zero hand-verified facts at publish time
- [ ] The sandbox deploy is keyed on apply-tags, and a red advisory job cannot suppress it (proven by construction of the needs graph)
- [ ] docs-freeze on an already-cut version is a green no-op; a prerelease tag moves neither `:latest` nor `/docs/latest/` nor publishes a non-prerelease Release
- [ ] Re-running the pipeline after a mid-run failure completes cleanly (idempotence per leg documented in the workflow header)

rubentalstra:
Closes #2776. Closes #2771. (Phase 2 of #2772; the approved design is `docs/plans/github-workflows-greenfield.md` §2.2.)

ONE workflow now triggers on `v*`: the rebuilt `release.yml` — `plan` (the full release-fact guard tier re-run against the tagged commit, mutation-proven per guard: a wrong version/appVersion/CITATION/changelog each fires before anything builds) → the draft GitHub Release + SLSA-L3 binaries → the three deferred-tag image lanes → the new reusable `scan-and-tag.yml` (shared with containers.yml's develop pushes — one implementation, and its completion IS "the tags exist") → in parallel the chart leg (`needs: scan-and-tag`; the poll loop, its job-name coupling, and its three distinct failures across v4.0.3–v4.0.5 — refuse-overwrite, a transient TLS death, waiting on a pre-tag signal (`manifest unknown`, run 32951186365) — are deleted), the sandbox (called workflow keyed on the applied tags, never a run conclusion; expected version as an input), and the prerelease-guarded, IDEMPOTENT docs freeze (proven: a second cut-version run on a frozen version exits 0 no-op; the old script exits 1) → `announce` (simulated against every needs-result combination; the simulation found and fixed a real printf defect).

Estate-wide grep proofs: exactly one `tags:` trigger; zero API polling of other workflows; zero `workflow_run` triggers. Deliberate behaviour changes, each adjudicated on the issue thread: any-dash prerelease semantics for the chart annotation; refuse-overwrite informational on the dry-run path only (run 32960523630); dispatching Release requires dispatching AT the tag (`plan` refuses with the exact command — the checkout-resolving lanes would otherwise build develop's tree under the tag's name); release-time `scan-and-tag` is strict (a skipped image blocks tagging/chart/sandbox); the smokes ride along from containers.yml (dropping them would have silently removed two release-time gates). `plan` deliberately does NOT run the helm goldens — the chart leg runs `validate.sh` itself before anything pushes, and plan stays toolchain-light.

The second commit updates every stale record of the old flow, including the supply-chain book page whose signer table named the CALLER workflows — the Fulcio certificate names the reusable workflow that owns the build (`build-image.yml`/`build-chart.yml`), which the chart lane's own read-back regexp already proved.

Gates: zizmor zero findings both sides (one ignore FEWER — the workflow_run dangerous-triggers ignore went with the trigger), actionlint clean, shellcheck, image-labels extended to release.yml + mutation-proven, docs-claims --all, changelog-structure, compose/statement/zenodo checks all green. The live tag proof is phase 6 (#2780) by design — first watch-items there: job-level `concurrency` on the reusable chart call, and `gh workflow run` from docs-freeze.


rubentalstra:
The four GitHub Advanced Security alerts (CodeQL `actions` cache-poisoning, alerts 44–47) are addressed by hardening rather than dismissal: every downstream job now checks out `github.sha` — provably the tag's commit, because `plan` refuses any dispatch not made AT the tag before anything runs — so the `workflow_dispatch` input no longer flows into any checkout ref. The flagged jobs also write no cache (`save-if: "false"`; the chart lane restores none at all, #2011), but removing the tainted ref is the honest fix and the alerts should close themselves on the next analysis. The branch is also synced with develop.

Issue #2236, opened 2026-08-10
test(terminology): run the SNOMED probe to completion, against the real NL release


rubentalstra:
**Retargeted 2026-09-03 (hold lifted):** the server this probe drives is FerroTERM, not Snowstorm. The hold existed because a Snowstorm + Elasticsearch import of the NL edition is beyond the development machine (about 8 GB of memory, an hour-plus import). FerroTERM builds the same RF2 zip into a 591 MB index in under a minute with `ferroterm-build` and opens it in half a second, so the run fits a laptop. The family (`scripts/deploy-probes/terminology.sh`) is rewritten to compose the FerroTERM image: a one-shot `build` service over the mounted release zip, then the server over the index; `P-TERM-UP`, `P-TERM-IMPORT` (now the build step), `P-TERM-SUBSUMES`, `P-TERM-VALIDATE` and `P-TERM-BROKEN` keep their meaning. The International-Edition question is answered by FerroTERM's RF2 loader rather than by Snowstorm's branch model, and `docs/conformance/deployment/README.md` is rewritten with it. Licensed content stays out of the tree, CI and images exactly as before. Now a sub-issue of #3085.

Split out of #2178 and moved off v3.17.4: the terminology family is **built and merged**, but has never been watched through to the end, because a full Snowstorm import takes longer than every other probe family combined and was holding up the milestone.

## What already works, verified

- The NL release is wired end to end: `SnomedCT_ManagedServiceNL_PRODUCTION_NL1000146_20260630T120000Z.zip`, 681 MB, MD5 `876355868299a8c4d1534e53de6e75a5` — **checked against the published value and matching**.
- The archive lives in the repository root and is gitignored (`SnomedCT_*.zip`), so the setup is "drop the download beside the checkout".
- The family is **opt-in** (`PROBE_ONLY=terminology`) — a default run skips it and says why, verified with the archive present.
- `P-TERM-UP` and `P-TERM-IMPORT` both reached, with Snowstorm serving FHIR and actively writing concepts, descriptions and relationships in batches. The licensed content genuinely loads.

## What has not been observed

The run was stopped during the import. So these are written and unexercised:

- `P-TERM-SUBSUMES` — `73211009` subsuming `44054006` over a real hierarchy, which is the whole reason a seeded code system was refused
- `P-TERM-VALIDATE` — the CDR validating a code through the configured provider rather than its in-process bundle
[...]

PR #2822, opened 2026-08-26
perf(docker): the build context drops 807 MB of material the compile never reads

### LINKED ISSUE #2813: perf(docker): the server image build context still carries 944 MB it never reads
Measured while landing #2811: fixing `.dockerignore` for the moved corpus cut the image build context from 1.17 GB to 944 MB — and the remaining 944 MB is still mostly material the `-p ferroehr-server` build never reads: `crates/*/vendor` (codegen inputs), `crates/openehr-term/assets` is read (embedded terminology — verify before excluding), `website/`, `docs/`, `docs-dist/`. Every gigabyte of context is transferred to the daemon on every image build, cold or warm.

Adjudicate each big tree first-hand against what the Dockerfile stages actually COPY/read (the build compiles the workspace, so crate sources and their include!-ed assets stay in; pure documentation/site/vendored-input trees go), then extend `.dockerignore` with the proof per entry. Measure before/after with the context-transfer line from a build.

## Acceptance criteria
- [ ] Each excluded tree carries its verified-unread reasoning; anything the compile reads stays
- [ ] The context size is measured before/after in the PR; the image builds and the compose smoke passes

rubentalstra:
Closes #2813.

Every image build transfers the whole context to the daemon, and 944 MB of it was material the `-p ferroehr-server` compile never reads. Measured honestly — a fresh buildx builder each time, so the transfer is full, not the context-cache increment:

| | context transfer |
|---|---|
| before | **944.63 MB** (18.0s) |
| after | **137.55 MB** (4.4s) |

Adjudication per entry, each verified against what the compile actually reads:

- **`fuzz/`** — its own nightly workspace, deliberately not a root-workspace member (`members = ["crates/*", "app/*", "tools/*"]`), so the build never parses it; `fuzz/corpus` alone is ~46 MB.
- **`website/` + `_site/`** — prose/HTML sources and the locally generated site output (22 + 36 MB); no `include_str!` or build.rs path reaches either.
- **`crates/*/vendor/` + `tools/openehr-codegen/vendor/`** — codegen inputs; the generated code is committed and the generator does not run in the image build. Every `include_str!` under `crates/` that names a `vendor/` path sits in `tests/` (the openehr_sdk corpus in `openehr-its/tests/it`). openehr-lang's vendored `.g4` grammars are reference text for the hand-written lexer, not a compile input.
- **`crates/openehr-its/schemas/xml/`** — read by the `xml_xsd_validity` test at runtime, never by the lib compile. **`schemas/json/` stays**: `src/json.rs` `include_str!`s `openehr_rm_1.1.0_all.json` into the lib, and the crate's packaged `include` list names it.
- **`*.zip`** — the single largest contributor: the untracked 681 MB SNOMED NL release parked in the repo root for #2236. No tracked zip exists and no build input is an archive.

Acceptance evidence (from-source image build + compose smoke) follows as a comment before merge — the cold in-image compile is running now.

`no-changelog`: build-context hygiene; no shipped byte changes.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -31,6 +31,37 @@ docs-dist/
 corpus/
 crates/openehr-adl/tests/corpus/
 
+# The fuzz harnesses (#2813): their own NIGHTLY workspace, deliberately not a
+# root-workspace member, so `-p ferroehr-server` never parses or compiles them
+# — and fuzz/corpus alone is ~46 MB of seed inputs.
+fuzz/
+
+# The website sources + the locally generated site build output (#2813): pure
+# prose/HTML, no crate reads them (verified: no include_str!/build.rs path
+# reaches website/ or _site/).
+website/
+_site/
+
+# Codegen INPUTS and reference grammars (#2813): the generated code is
+# committed, the generator does not run in the image build, and every
+# include_str! under crates/ that reaches a vendor/ path sits in tests/ (the
+# openehr_sdk corpus in openehr-its/tests/it). openehr-lang's vendored .g4
+# grammar set is reference text for the hand-written lexer, not a compile
+# input (no build.rs, no include_str! names it).
+crates/*/vendor/
+tools/openehr-codegen/vendor/
+
+# The vendored XSD bundles (#2813): read by the xml_xsd_validity test at
+# RUNTIME, never by the lib compile. schemas/json/ STAYS — src/json.rs
+# include_str!s openehr_rm_1.1.0_all.json into the lib (it is even in the
+# crate's packaged `include` list).
+crates/openehr-its/schemas/xml/
+
+# Local terminology releases (#2813): the 681 MB SNOMED NL zip parked in the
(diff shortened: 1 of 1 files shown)


rubentalstra:
Acceptance evidence, run live on the slimmed context:

- **From-source image build**: `docker compose build ferroehr` completed — `naming to ghcr.io/rubentalstra/ferroehr:local` / `Built` — proving no compile input was excluded (the in-image build compiles the whole workspace path for `-p ferroehr-server`, including the lib-embedded `schemas/json/openehr_rm_1.1.0_all.json`).
- **Compose smoke**: `docker compose up -d --wait ferroehr` → postgres Healthy, ferroehr Healthy. `/health/readiness` → `{"status":"UP","components":{"audit_sender":{"status":"UP"},"db":{"status":"UP"},"migrations":{"status":"UP"}}}`; an authed `GET /definition/template/adl1.4` → 200. Stack torn down after.
- Context transfer (fresh buildx builder each side): 944.63 MB → 137.55 MB.

Issue #2864, opened 2026-08-27
chore(conformance): the Veredictum pin lags three releases — bump to 0.1.0-alpha.6 with its own acceptance run


rubentalstra:
Found during #2861: `scripts/lib/veredictum.sh` pins the CNF runner at `0.1.0-alpha.2` while crates.io publishes through `0.1.0-alpha.6` (alpha.3, alpha.4 — the lib API — and alpha.6 unadopted). The pin's own contract applies: a bump moves the catalogue, the oracle and the verdict pipeline together, so it lands as its OWN change re-proven by a full `scripts/conformance.sh` acceptance run against the committed baseline — never riding another PR, where a verdict delta would be unattributable (the #2861 worker correctly refused exactly that).

- [ ] The pin moves to the newest published `veredictum` (one-line change in `scripts/lib/veredictum.sh`)
- [ ] A fresh `bash scripts/conformance.sh` acceptance run lands with the bump; any verdict delta is adjudicated instrument-vs-application per `.claude/rules/cnf-triage.md` before anything is committed
- [ ] The committed `docs/conformance/ferroehr/` record reflects the new instrument (this also clears the carried-rc2-measurement warning noted at #2752)

rubentalstra:
Superseded by the owner's direct ruling during #2861: the pin moved to `0.1.0-alpha.6` as its own commit on PR #2865 (both halves verified resolving — the binary reports alpha.6, the tag exists for the catalogue), rather than as this issue's standalone PR. What survives of this issue's contract is the acceptance-run half, and it lives on #2865's DO-NOT-MERGE gate: the deferred `scripts/conformance.sh` run must land green before that PR merges, and any verdict delta is attributed between the two candidate causes (the S3776 refactor vs the instrument bump) per `cnf-triage.md` — the pin commit is separately revertible precisely so the attribution can isolate it. Closing; the record continues on #2865.

Issue #2861, opened 2026-08-27
chore(quality): clear the 24 SonarCloud findings — the post-#2752 residue and its own new surface


rubentalstra:
The dashboard reads 24 open (quality gate: passed; Reliability A ×4, Maintainability A ×20) after the #2752 clearance merged. Enumerated live through the MCP:

- **4× `shelldre:S7688`** (`[` → `[[`), all in `scripts/release/publish-crates.sh:90,121,125,127` — the #2837 script, born after the shellcheck lane's sweep and never adjudicated.
- **20× `rust:S3776`** (cognitive complexity over 15): six created at the 2026-08-27 post-merge analysis (`ferroehr-rest/extensions/fhir.rs:408` cc25, `flat/webtemplate/builder_v2_4.rs:618` cc19, `openehr-its/tests/it/xml_xsd_validity.rs:258` cc32, `codegen/load/oas.rs:100`, `codegen/render/emit.rs:1549`, `codegen/render/emit_opt.rs:497`) — partly new surface the #2752 refactor itself produced — plus fourteen older sites the clearance missed at moved lines (`ferroehr-server/src/lib.rs:455` cc22, `openehr-adl` validate/convert/parse ×3, `codegen` testsupport ×3 + render ×4 + analyze ×1, the iso8601 template cc22, `openehr-its/tests/it/flat.rs:188`).

Same law as #2752: work the LIVE MCP list; refactor with the same care the code was written with (named, documented helpers); a linear match over a closed spec vocabulary that resists honest decomposition is MCP-accepted per site with the reason, never blanket-suppressed; behaviour pinned by the existing suites; generated-crate sites go through the emitter. Conformance zero-drift if engine-adjacent code moves.

## Acceptance criteria
- [x] The 4 shell findings fixed (the lane's shellcheck + a re-run of the script's verify path stay green)
- [x] The 20 S3776 sites refactored or per-site MCP-accepted with recorded reasons; dispositions registered here
- [x] The dashboard's open count reaches the standing-accepted set only

rubentalstra:
## `rust:S3776` dispositions — 20 of 20 refactored, 0 accepted

Worked from the live MCP list (`projectKeys=[rubentalstra_FerroEHR]`,
`issueStatuses=[OPEN,CONFIRMED]`), which returned the same 20 sites the body
enumerates. Every one decomposed; no site needed an acceptance, because in all
20 the complexity came from loop/branch NESTING that the code's own comments
[...]

PR #2865, opened 2026-08-27
refactor(quality): the twenty cognitive-complexity sites split into named helpers

rubentalstra:
> [!NOTE]
> The deferred conformance run landed on this branch (29f817f24): zero verdict drift on 0.0.43 + alpha.6 — the merge gate is satisfied.

Part of #2861.

SonarCloud reports 20 open `rust:S3776` findings — functions whose cognitive
complexity exceeds 15. All 20 are refactored here; none is accepted. Working
from the live MCP list rather than the issue's snapshot confirmed the same 20
sites, and in every one the complexity came from loop and branch NESTING that
the code's own comments already described as separate phases. Naming those
phases was the whole fix: no closed-vocabulary match surface turned up in this
batch, so there was nothing to re-adjudicate.

The four `shelldre:S7688` shell findings on the same issue were already fixed
and merged as #2863 and are untouched here; the MCP still reports them OPEN
only because that merge has not been re-analysed.

## What moved

**The application.** The FHIR mapping store's dispatcher was one `match` with
a nested `match` per arm — resolve the id, read the body, call the backend,
render. It becomes a five-arm route table over one small handler per HTTP
operation, each with the path it serves in its doc line. The boot sequence
sheds three phases it had already separated with comments: `assemble_service`
(the service and every optional collaborator it carries), `wire_authz` (the
RBAC/ABAC handle, `None` when authentication is off), and
`log_resolved_posture` (the per-subsystem startup lines). What is left is the
linear startup order, and its `too_many_lines` expectation stays with the
reason restated to say so.

**The spec crates.** The ADL specialisation checker hands each of its
production checks to a named function — owner resolution through a
differential path, the absent-parent-attribute case, the child pairing, and
the newly-added-object VSONIF rule, which now sits with its own citation
instead of five levels deep in a loop. The 1.4→2 conversion walk gives the
`C_ATTRIBUTE_TUPLE` half its own function. The archetype-slot parser splits
its two mutually exclusive bodies — the ADL2-only `closed` marker and the open
`matches { include … exclude … }` block — and returns a `SlotBody` value
rather than threading four mutable locals through an if/else. The Simplified
Formats web-template builder reads one tuple row through `tuple_row_unit`.

**The ISO 8601 duration parser** is a generation-twin template, so the change
was made in `tools/openehr-codegen/templates/…/iso8601_parse.rs` and `emit`
restamped the `v1_2` and `v1_3` copies. Its loop state becomes a
`DurationScan` (whose `last_slot` starts at `-1` in a hand-written `Default`,
per the inline-default rule) and its digit arm becomes
`read_duration_component`.

**The generator.** Both least-fixpoint computations (`base_closure`,
`constructible_classes`) become `while <one named round>() {}`, which is what
they always were. The struct renderer sheds its prologue and its
`// inherited:` group header; the enum renderer sheds its two import
computations; both share a new `generic_decl`. `emit_enum_conversions` splits
into its int and string halves — byte-identical output, verified by the drift
check. `emit_structural_file` separates collecting the `_type` dispatch
inventory from writing the file. `element_field` splits its container wrapping
and its wire default out. `model_query` gains `view_props` for the
declared-versus-flattened choice.

**The test harnesses.** The three `testsupport` projections nest one loop level
per named function, so a gap report reads as "per unit, per class" rather than
five loops deep. The two integration tests gain named helpers — a
`RoundTripTally` for the FLAT sweep and an element-folding pass for the XSD
inventory — with every assertion, count and threshold intact. No test was
weakened, skipped or deleted.

## Crate versions

Entries in `openehr-adl`, `openehr-its` and `openehr-base` change packaged
content, so the eight `openehr-*` crates step to `0.0.43` in lockstep, with
both lockfiles refreshed and a changelog entry for the step. The generated
output is unchanged apart from the two restamped template copies.

## Gates

Run and green:

- `cargo fmt --all` — clean
- `cargo clippy --workspace --exclude ferroehr-admin-ui --all-targets --all-features -- -D warnings` — clean
- `cargo nextest run --workspace` — **4859 passed, 0 failed, 6 skipped**
- `RUSTDOCFLAGS="-D warnings" cargo doc … --document-private-items` — clean
- `scripts/checks/codegen-drift.sh` — clean, which is what proves the template restamp is byte-exact
- `comment-style`, `typed-status`, `default-style`, `error-source-chain`, `shellcheck-lane` — all OK

Outstanding:

- **`bash scripts/conformance.sh` — NOT RUN (owner-deferred).** It must land green,
  with zero verdict drift, before this merges. It also has to clear the pin step
  described above, since that moves the catalogue and the oracle together.

## The CNF runner pin

`scripts/lib/veredictum.sh` moves from `0.1.0-alpha.2` to `0.1.0-alpha.6` in its
own commit, per an explicit owner instruction to run the newest published
version. crates.io publishes through alpha.6, so the pin was three releases
behind (alpha.3, alpha.4, alpha.6 unadopted; alpha.5 never published). Both
halves resolve — the binary reports `veredictum 0.1.0-alpha.6`, and
`refs/tags/v0.1.0-alpha.6` exists for the catalogue checkout.

This supersedes the "stay on alpha.2, bump separately under #2864" plan I had
argued for and that #2864 records; the owner overrode it. The reason I argued
for the split still stands as a *review* obligation rather than a scoping one:
the pin's own comment requires a bump to be re-proven by a full conformance run,
and bumping the instrument in the same cycle as a refactor means a verdict delta
is ambiguous between them. So the deferred run must attribute any delta
explicitly, per `.claude/rules/cnf-triage.md`, before either change is accepted.
It is kept in a separate commit precisely so it can be reverted alone if that
attribution goes against it.

Not `Closes` — the issue's final criterion is the dashboard count, which only
the post-merge analysis can confirm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,14 @@ workflow refuses a tag that has no matching section here.
 
 ## [Unreleased]
 
+### Changed
+
+- The `openehr-*` spec crates step to `0.0.43`. The change is internal
+  structure only: over-complex functions in the ADL engine, the Simplified
+  Formats web-template builder and the ISO 8601 duration parser were split
+  into named, documented helpers. The wire behaviour, the generated output and
+  every gate are unchanged, and nothing an API consumer calls moved.
+
 ## [4.0.8] - 2026-08-27
 
 ### Changed
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5426,7 +5426,7 @@ checksum = "c08d65885ee38876c4f86fa503fb49d7b507c2b62552df7c70b2fce627e06381"
 
 [[package]]
 name = "openehr-adl"
-version = "0.0.42"
+version = "0.0.43"
 dependencies = [
  "indexmap 2.14.0",
  "openehr-am",
@@ -5442,7 +5442,7 @@ dependencies = [
 
 [[package]]
 name = "openehr-am"
-version = "0.0.42"
+version = "0.0.43"
 dependencies = [
  "openehr-base",
  "openehr-lang",
@@ -5452,7 +5452,7 @@ dependencies = [
 
 [[package]]
 name = "openehr-base"
-version = "0.0.42"
+version = "0.0.43"
 dependencies = [
  "serde",
  "serde_json",
@@ -5471,7 +5471,7 @@ dependencies = [
 
 [[package]]
 name = "openehr-its"
-version = "0.0.42"
+version = "0.0.43"
 dependencies = [
  "async-trait",
  "axum",
@@ -5500,7 +5500,7 @@ dependencies = [
 
 [[package]]
 name = "openehr-lang"
-version = "0.0.42"
+version = "0.0.43"
 dependencies = [
  "assert_fs",
  "chumsky",
@@ -5514,7 +5514,7 @@ dependencies = [
 
 [[package]]
 name = "openehr-query"
-version = "0.0.42"
+version = "0.0.43"
 dependencies = [
  "chumsky",
  "logos",
@@ -5523,7 +5523,7 @@ dependencies = [
 
 [[package]]
 name = "openehr-rm"
-version = "0.0.42"
+version = "0.0.43"
 dependencies = [
  "insta",
  "jsonschema",
@@ -5541,7 +5541,7 @@ dependencies = [
 
 [[package]]
 name = "openehr-term"
-version = "0.0.42"
+version = "0.0.43"
 dependencies = [
  "openehr-base",
  "roxmltree",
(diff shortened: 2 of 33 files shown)


rubentalstra:
Merge blocker (posted so it survives a body edit): `bash scripts/conformance.sh` has NOT been run for this branch — it was owner-deferred when the machine was needed mid-build. Do not merge until it lands with zero verdict drift, and until any delta is attributed between the two candidate causes in this branch (the refactor, and the `0.1.0-alpha.2` → `0.1.0-alpha.6` CNF runner pin step) per `.claude/rules/cnf-triage.md`. The pin commit is deliberately separate so it can be reverted alone.

rubentalstra:
The deferred conformance run has landed, on this branch (commit 29f817f24): **1102 case-records — 1064 passed / 0 failed / 0 errored / 38 n-a; `verdicts.json` and all six badges byte-identical to the committed baseline.** Both candidate causes are thereby cleared at once — the twenty S3776 refactors and the alpha.6 instrument bump produced zero observed-behaviour change, so no attribution question arises. The only artifact movement is identity truth (SUT label 4.0.6-rc2 → 4.0.8; runner alpha.2 → alpha.6), which is the pin commit's own acceptance record riding the PR that carries the pin. The DO-NOT-MERGE gate is satisfied.

PR #2903, opened 2026-08-29
feat(aql): FOLDER CONTAINS FOLDER unites the by-value subtree with items-referenced versioned folders

### LINKED ISSUE #2887: AQL: adjudicate FOLDER CONTAINS FOLDER over items-referenced VERSIONED_FOLDERs
`FOLDER.items` is `List<OBJECT_REF>` and an OBJECT_REF may name a `VERSIONED_FOLDER` (RM common master05 — the directory package constrains the ref target type nowhere). The #2880 edge classifier renders `FOLDER f1 CONTAINS FOLDER f2` as the by-value `folders` sub-tree (a strict nested-set descendant inside one versioned object) and `FOLDER CONTAINS <other VO>` as the `items` reference resolution. A folder whose `items` references a separate VERSIONED_FOLDER is therefore not matched by `FOLDER CONTAINS FOLDER` today.

Whether it SHOULD be is a spec-silence question: QUERY master03 §Containment defines CONTAINS as the parent/child relationship and says nothing about reference edges at all — the items resolution for compositions is itself an RM-grounded interpretation. Uniting both edges for a FOLDER child (sub-tree OR items-referenced versioned folder) is mechanical once adjudicated, but doing it speculatively risks wrong duplicates and an unpinned envelope.

## Acceptance criteria

- [ ] Adjudicate first-hand from the vendored QUERY + RM texts whether `FOLDER CONTAINS FOLDER` covers items-referenced versioned folders; record spec silence in Veredictum's ambiguity register with a typed disposition.
- [ ] If in: the classifier renders the union edge, planner + integration tests pin it (a versioned folder referenced from another directory's items).
- [ ] If out: the boundary lands in the book's AQL envelope section and a refusal/absence test pins it.

rubentalstra:
Closes #2887.

**Adjudication (first-hand from the vendored texts): IN, one reference hop.** QUERY master03 §Containment defines `CONTAINS` as the parent/child hierarchy and says nothing about reference edges; RM common master05 declares `FOLDER.items: List<OBJECT_REF>` as "references to other (usually) versioned objects **logically in this folder**" with no target-type restriction, and a `VERSIONED_FOLDER`'s every version "is a Folder structure". The same RM sentence that grounds `FOLDER CONTAINS COMPOSITION`'s items resolution therefore grounds the folder-typed target: refusing exactly the parallel edge for `VERSIONED_FOLDER` targets would make the envelope asymmetric on no spec ground. The hop is NOT transitive across further referenced versioned folders — that closure is cycle-prone (folder A referencing B referencing A) and unbounded in SQL, and the chain is expressible explicitly (`f1 CONTAINS f2 CONTAINS f3`), each edge evaluated pairwise.

**The engine change** (`app/ferroehr/src/aql/sql/from.rs`): the `FOLDER CONTAINS FOLDER` edge (`EdgeKind::FolderSubtree` → `FolderChild`) becomes the union — a by-value STRICT nested-set descendant in the parent's own tree, OR any folder row of an items-referenced `VERSIONED_FOLDER` (the existing `folder_items_exists` correlation over the parent's subtree). The child now opens its own version group in both branches (in the by-value branch its spine row IS the parent's, so latest/ALL_VERSIONS scoping is unchanged); the same union renders in the OR/NOT-CONTAINS anchor correlation. Streaming is untouched — folder edges were already flat-shape-only.

**Pinned by:** the planner test (`folder_contains_folder_unites_subtree_and_items_reference`: strict interval + OR + items lookup in the SQL) and a real-database integration test (`folder_contains_folder_reaches_items_referenced_versioned_folders`): a directory whose sub-folder references a second VERSIONED_FOLDER committed through a CONTRIBUTION — the exact six-pair union set (no duplicates, reference root + its by-value descendants, one hop only), plus the chained `f1 CONTAINS f2 CONTAINS COMPOSITION c` hop across the reference. The #2888 suite still passes unchanged.

Book envelope updated (querying-aql.md); changelog entry added. Conformance zero-drift run quoted in a comment before merge (the aql-engine rule).

The ambiguity-register entry (spec silence, disposition: RM-grounded acceptance, one hop) is Veredictum-side material — the ready-to-land entry text is on the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ workflow refuses a tag that has no matching section here.
 
 ### Changed
 
+- AQL `FOLDER f1 CONTAINS FOLDER f2` now also matches the folders of a
+  `VERSIONED_FOLDER` that a folder's `items` reference — the same
+  reference-resolution ground as `FOLDER CONTAINS COMPOSITION`, one
+  reference hop (chain `CONTAINS` to follow further hops). Strict
+  sub-folder matching is unchanged.
 - The Helm chart's committed `appVersion` now always equals the product
   version, bumped in every release PR beside the compose image tags (chart
   6.0.27). Published charts are unchanged — the release pipeline keeps
--- a/app/ferroehr/src/aql/sql/from.rs
+++ b/app/ferroehr/src/aql/sql/from.rs
@@ -55,10 +55,14 @@ enum EdgeKind {
     /// common master05 — folders hold *references* to versioned objects; RM
     /// ehr master04 §Folders). The child opens its own version group.
     FolderItems,
-    /// `FOLDER CONTAINS FOLDER`: the by-value `folders` nesting inside one
-    /// versioned folder tree — a STRICT descendant interval, or the same-typed
-    /// pair matches itself.
-    FolderSubtree,
+    /// `FOLDER CONTAINS FOLDER`: the union of the by-value `folders` nesting
+    /// (a STRICT descendant interval — the inclusive join would self-match
+    /// every folder row) and an items-referenced `VERSIONED_FOLDER`'s folder
+    /// rows (RM common master05: `items` references name versioned objects
+    /// "logically in this folder", with no target-type restriction — the same
+    /// ground as `FolderItems`; one reference hop, adjudicated on #2887). The
+    /// child opens its own version group either way.
+    FolderChild,
 }
 
 /// Classifies a `CONTAINS` edge, or refuses a pair the RM defines no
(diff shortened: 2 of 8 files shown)


rubentalstra:
Zero-drift evidence (the aql-engine rule): two full `scripts/conformance.sh` runs against this branch's SUT — **1102 case-records: 1064 passed / 0 failed / 0 errored / 38 n-a**, 43 capability verdicts, `verdicts.json` byte-identical to the committed baseline. The second run's docker build was fully CACHED including the workspace compile step, proving both runs built the identical (this-branch) tree. The only artifact motion is the SUT version stamp (4.0.8 → 4.0.9 — the committed baseline predated the v4.0.9 cut), committed here.

Issue #2916, opened 2026-08-29
perf(rest): the community read benchmark's open challenge — GET a stored composition in under 1 ms


rubentalstra:
The community benchmark thread (<https://discourse.openehr.org/t/17224>) closed its round with a direct challenge (borut.jures, 2026-08-25): 'Can you get your read speed under 1 ms?' — the current cross-machine-normalized read is ~1.3-1.7 ms/query over the 7-endpoint GET mix (700k reads over 100k stored Vital-signs compositions), with ArcEHR at ~0.27 ms as the pace-setter. The write side is effectively tied after normalization.

Profile-first (the standing rule; the /flamegraph skill): the #17224 fix already proved PostgreSQL does <0.1 ms of work per read, so the remaining budget lives in our HTTP/auth/audit/serialization envelope. Known suspects to measure, not guess: per-request ATNA audit record insertion (measure the with/without delta and whether the queue amortizes it), Basic-auth Argon2 verification per request (the verified-credential cache's hit path), response compression level, the axum middleware stack per-request cost, and jsonb detoast/decompression of the body column (couples to #2913 — a text column changes this too).

Constraints: conformance stays green (never trade wire correctness), ATNA + signing stay ON in any published number (the defaults are the product), and the measured protocol is the thread's own harness (their src.zip) so the numbers stay comparable. Publish the outcome on the thread; #2744's community-harness re-run rides the same work.

## Acceptance criteria
- [ ] A flamegraph-backed cost breakdown of one GET (numbers per layer, committed to the issue)
- [ ] Every accepted lever lands with its measured delta; every rejected one carries the measured reason
- [ ] The 7-endpoint read mix re-measured on the thread's harness with defaults ON; result posted to the thread (one short comment)

rubentalstra:
Moved to v4.0.12 (owner decision, 2026-08-29): the measurement instrument for this challenge is the Veredictum benchmark tool (its tracker's universal-benchmark work), not the community thread's src harness — the re-measure acceptance criterion reads against that tool once it ships. The profile-first cost breakdown still lands here when picked up.

rubentalstra:
Closed as not planned (owner decision, 2026-08-31, with #2744): the community thread's harness is retired as a measurement instrument on our side — performance measurement happens on the Veredictum benchmark tool once it ships. The sub-1 ms read challenge stays a real target, but chasing it gets filed fresh against that instrument's numbers when the tool is ready, not against this issue's harness-bound acceptance criteria.

PR #2926, opened 2026-08-29
fix(rest): reads serve the canonical bytes verbatim — the body column stops being jsonb

### LINKED ISSUE #2913: fix(rest): the stored-body fast path serves jsonb-reordered keys — the canonical field order is lost on every read
Observed live by the community (borut.jures, <https://discourse.openehr.org/t/17224>, 2026-08-25: 'your latest changes resulted in returning instances with unsorted RM properties') and confirmed in the schema: the #2698-era read optimization serves `vo_version.body` verbatim, and that column is **`jsonb`** (`app/ferroehr/migrations/ehr/0001_baseline.sql:419`) — PostgreSQL's jsonb stores object members in its own internal key order (length-then-bytes; the PostgreSQL JSON-types documentation states jsonb 'does not preserve … the order of object keys'). So a GET no longer serves the canonical encoding the codec produced at commit: `_type` is not first, and the BMM declaration order is gone.

No released openEHR text mandates member order on the JSON wire (RFC 8259 makes it insignificant), so this is not a conformance defect — the forum said as much. It IS an internal-contract defect: the repo's canonical-output contract pins `_type`-first + BMM field order for everything the codec emits, the fidelity gates prove that property on the CODEC while the READ path silently bypasses it, and a fetch→recommit round trip is no longer byte-stable. Streaming clients that dispatch on an early `_type` also pay.

Adjudicate, then fix at the storage layer rather than by re-encoding at read (which would resurrect the row-rebuild cost the optimization removed): the leading candidate is storing the canonical body as `text COMPRESSION lz4` so the served bytes ARE the commit-time canonical encoding, with every SQL site that inspects `body` as jsonb (the AQL whole-object assembly, dump/load, the spec-profile stamp assessment) audited for a cast or a better read. Greenfield migrations are edited in place (#2452 ruling).

## Acceptance criteria
- [ ] A GET of any stored versioned object serves byte-identical canonical JSON to what the codec emitted at commit (`_type` first, BMM field order) — pinned by a wire test that commits via the codec and compares the GET body BYTES, not the parsed value
- [ ] The server-stamped `uid` sits at its BMM position in the stored text — the pinning test commits a body WITHOUT a `uid`, so the stamped-field placement is what the byte comparison exercises (found by borutjures below)
- [ ] AQL whole-object projections serve the same canonical bytes; every jsonb-consuming site of `body` is dispositioned (cast, changed read, or removed)
- [ ] The #17224 read-latency envelope is re-measured after the change and quoted (no regression beyond noise)
- [ ] The adjudication (spec-silent; our own canonical-output contract is the ground) is recorded in the fixing PR with the RFC 8259 + PostgreSQL doc citations

rubentalstra:
Closes #2913.

**The fix, at the layer the defect lived:** `vo_version.body` becomes `text COMPRESSION lz4` (greenfield baseline edit; the cold mirror follows via `LIKE`), and the stored bytes are the accepted, uid-stamped canonical value serialized BEFORE node decomposition — so a point read serves the codec's `_type`-first, BMM-declared field order verbatim instead of PostgreSQL's jsonb key ordering (length-then-bytes), which is what the community observed on the wire.

**The stamped uid lands at its BMM position** (the borutjures refinement): `stamp_version_uid` now places a fresh `uid` after `archetype_node_id` (the RM BMM declares LOCATABLE as name, archetype_node_id, uid, links) instead of appending to the map's end; a client-supplied `uid` keeps its position by in-place replace.

**Signature semantics unchanged:** the `VERSION.signature` canonicalization is RFC 8785 (key-order-insensitive), so signing the node-row reassembly still proves the stored and served bytes; verified in place with the doc updated.

**Every body consumer dispositioned:**
- The typed read path parses the stored text (order kept via `preserve_order`); the raw JSON-accept passthrough serves the column verbatim — that IS the fix's serving path.
- The rare SQL that inspects the body casts (`(body)::jsonb #>>/->>`): the template/category lookups in `meta.rs`/`placement.rs`, the node-repo category probe, the tag-category filter.
- AQL whole-object assembly (`read_root_bodies`) parses the text — whole-object cells now carry canonical order too.
- **Dump/export reads the stored bytes instead of reassembling from node rows** (`stored_body_all`, tier-spanning): the node `data` fragments are jsonb and leaked their ordering into archives — the export/load round-trip is byte-equal again, pinned by the existing round-trip suite. The storage-parity sweep keeps the node-reassembly reader (comparing the two is its whole point).
- Import/archive-load bind serialized text; `StorageError::BodyDecode` names the (corruption-only) parse failure, mapped to the codec-fault 500.

**Pinned by:** the rewritten `composition_get_serves_stored_json_verbatim` — a create carrying NO body `uid`, whose GET must open with `_type`, carry `["_type","name","archetype_node_id","uid"]` as the leading keys, and be **byte-identical to the canonical codec's own re-encoding** of itself, with the version-addressed read serving the same bytes. (The old test pinned the jsonb artifact — the body opening with `uid` — and is replaced by the strictly stronger contract, per the adjudicated semantics change.)

Suites: ferroehr 989/989, ferroehr-rest 773/773; fmt/clippy/comment-style/typed-status clean. Zero-drift conformance run + the read-latency sanity check land as comments before merge. Changelog `### Fixed` entry included (wire-visible).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,16 @@ workflow refuses a tag that has no matching section here.
 
 ## [Unreleased]
 
+### Fixed
+
+- Reads serve the exact canonical bytes the commit accepted: `_type` first
+  and every field at its spec-declared position — including the
+  server-stamped `uid`, which now lands at its place in the document instead
+  of the end. Previously the stored body passed through a `jsonb` column
+  and PostgreSQL's own key ordering leaked onto the wire (reported by the
+  community). Dump archives carry the same faithful bytes, so an
+  export/load round-trip is byte-equal again.
+
 ## [4.0.10] - 2026-08-29
 
 ### Added
--- a/app/ferroehr-rest/tests/it/protocol_tail.rs
+++ b/app/ferroehr-rest/tests/it/protocol_tail.rs
@@ -115,6 +115,46 @@ fn vo_of(ovid: &str) -> &str {
     ovid.split("::").next().expect("vo uuid")
 }
 
+/// [`commit_ips_composition`] with a caller-supplied composition body.
+async fn commit_composition_body(app: &Router, body: Value) -> (String, String) {
+    let (status, h, _b) = send(
+        app,
+        Request::builder()
+            .method("POST")
+            .uri(format!("{BASE}/ehr"))
+            .body(Body::empty())
+            .unwrap(),
+    )
+    .await;
+    assert_eq!(status, StatusCode::CREATED);
+    let ehr_id = etag_uid(&h);
+
+    let (status, _h, resp) = send(
+        app,
+        Request::builder()
+            .method("POST")
+            .uri(format!("{BASE}/definition/template/adl1.4"))
+            .header(header::CONTENT_TYPE, "application/xml")
+            .body(Body::from(opt_xml()))
+            .unwrap(),
+    )
+    .await;
+    assert_eq!(status, StatusCode::CREATED, "OPT upload: {resp}");
+
+    let (status, h, resp) = send(
+        app,
+        Request::builder()
+            .method("POST")
+            .uri(format!("{BASE}/ehr/{ehr_id}/composition"))
+            .header(header::CONTENT_TYPE, "application/json")
+            .body(Body::from(body.to_string()))
+            .unwrap(),
+    )
+    .await;
+    assert_eq!(status, StatusCode::CREATED, "composition commit: {resp}");
+    (ehr_id, etag_uid(&h))
+}
+
 /// Create an EHR, upload the IPS OPT, and commit the IPS composition; return the
 /// `(ehr_id, version_uid)` of the committed COMPOSITION.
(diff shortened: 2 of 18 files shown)


rubentalstra:
Zero-drift conformance run on this branch: 1102 case-records, 1064 passed / 0 failed / 0 errored / 38 n-a — identical verdicts to the committed baseline; the only artifact motion is the product-version stamp (4.0.9 → 4.0.10), committed here as the restamped baseline.

Read-latency re-measure (the #2916-shape single-client keep-alive GET loop, 2000 reads of one stored composition, quickstart compose on this dev box): fix image 2.72 ms and 6.77 ms per GET across two runs, pre-fix 4.0.10 image 4.60 ms. Run-to-run noise on this machine spans the delta in both directions, so the honest statement is: no measurable regression, and the JSON fast path now serves the stored bytes without a decode + re-encode. The conformance perf classes remain the real instrument.

Issue #2927, opened 2026-08-29
fix(aql): lower path-embedded predicates on fragment steps and whole-object roots — currently typed rejects


rubentalstra:
Verified empirically against the planner (`build_sql`, 2026-08-29, on the #2919 branch): two path-embedded predicate positions never reach the generated SQL — the query builds and answers as if the predicate were not written, the silent-wrong-answer class the engine's own hard rule forbids.

1. **A predicate on the path ROOT** (`c[openEHR-EHR-COMPOSITION.report.v1]/name/value = 'x'` in WHERE/SELECT): the emitted SQL carries only `rm_type IN (...)` and the comparison — no archetype condition. Any composition's name matches.
2. **A predicate on a FRAGMENT step** (`c/links[at0001]/target/value = 'x'`): the fragment jsonpath renders `$.links.target.value` with the constraint discarded (`fragment_jsonpath` writes only step names).

Anchor-step predicates (`data[at0001]`, `events[at0002]`, …) are unaffected — `anchored_walk`/`whole_object_alias` lower them via `node_constraint_conds`.

The #2919 branch lands the tractable half: the root predicate lowers for LEAF reads (conditions on the source node alias — CASE-guarded in scalar/projection contexts, conjoined in existential subqueries), and the two remaining shapes become TYPED REJECTS (a fragment-step predicate anywhere; a root predicate on a whole-object projection). This issue tracks the real lowering of the rejected shapes.

QUERY master03 §Identified Paths defines node predicates at any path position; no released text scopes them to structure nodes — so the fragment-step form is a legitimate query the engine must eventually answer, presumably via jsonpath filter expressions with `vars`-bound literals (`jsonb_path_query(data, '$.links[*] ? (@.archetype_node_id == $a).target.value', vars)`), which keeps binding parameterized.

## Acceptance criteria
- [ ] A fragment-step predicate (archetype/name/standard forms) lowers to a jsonpath filter (or an equivalently exact shape) with literals bound via `vars`, never string-spliced
- [ ] A root predicate on a whole-object projection has a decided lowering (NULL cell via guarded locators, or a documented reject)
- [ ] The #2919-era typed rejects are replaced by the lowering, and their reject tests flipped to accepted-shape pins
- [ ] Planner + DB tests over data where the predicate selects the non-first element

Issue #1448, opened 2026-08-01
fix(aql): unify LIKE/matches lowering on anchored multi-valued paths with the existential form


rubentalstra:
`app/ferroehr/src/aql/sql/predicate.rs:90`: LIKE/matches on anchored multi-valued paths still lower through the scalar LIMIT-1 extraction, whose matched-node choice is order-undefined when a path matches several nodes — the comparison operators already use the existential lowering (exists_compare). AQL 1.1 semantics require a defined answer; scheduling per the owner (QUERY program closed — candidate for the current patch cycle).

(Backing issue for a code `TODO` per the owner rule 2026-08-01 — every marker references its tracker issue; sweep #1432.)

## Acceptance criteria
- [ ] LIKE/matches use the existential lowering on multi-valued paths; corpus + CNF query cases pin it

PR #2928, opened 2026-08-29
fix(aql): predicates and projections see every element of a list-valued fragment path

### LINKED ISSUE #2919: fix(aql): predicates and projections over multi-valued fragment paths see only the first array element
A path step that resolves to a non-structure RM type is addressed inside the anchor node's `data` jsonb, and that extraction is `jsonb_path_query_first` unconditionally (`app/ferroehr/src/aql/sql/expr.rs:49`). PostgreSQL's lax mode auto-unwraps arrays, so the path yields N items and `_first` returns item 1. Every list-valued fragment attribute is therefore invisible past its first element: `LOCATABLE.links`, `EVENT_CONTEXT.participations`, `ENTRY.other_participations`, `PARTY_IDENTIFIED.identifiers`, `DV_TEXT.mappings` (all `is_structure_root: false`).

`WHERE c/context/participations/performer/name = 'X'` returns no row when X is the second performer, and `SELECT i/links/target/value` returns one target where the data holds three. Nothing errors. Verified empirically against the live PG18 testkit: `jsonb_path_query_first('{"links":[{"target":{"value":"a"}},{"target":{"value":"b"}}]}', '$.links.target.value') → "a"`.

The engine already carries the right semantics for the other half of the same split: `data_leaf_exists` (`app/ferroehr/src/aql/sql/value.rs:137`) lowers a predicate on a multi-valued ANCHORED path existentially, with the adjudication recorded in place (QUERY master03 §WHERE is silent on multi-valued predicates — any-match is our documented decision). `LeafPath.multi_valued` is computed in `analyze.rs` and read by no SQL code, so the fragment half never reaches that decision. The fix extends the EXISTING in-repo adjudication to the fragment half — not a new semantic.

Discourse ties: <https://discourse.openehr.org/t/16797> (filtering on `links/target/value` — the exact breaking shape) and <https://discourse.openehr.org/t/16799>.

## Acceptance criteria
- [ ] A predicate whose leaf path crosses a multi-valued fragment step lowers existentially (`jsonb_path_exists` / an unnested `EXISTS`), never through `jsonb_path_query_first`
- [ ] `LeafPath.multi_valued` / the fragment-step flags are consumed by the SQL layer, or deleted if genuinely dead
- [ ] Projection over a multi-valued fragment step has a decided, documented behaviour (first element, an array cell, or a typed refusal) with a `// NOTE:` citing the spec silence
- [ ] Tests over stored data with ≥2 `links`, ≥2 `context/participations` and ≥2 `identifiers` where only the LAST element matches — for `=`, `LIKE`, `MATCHES` and `EXISTS`
- [ ] A CNF wire case for LINK filtering (extends the LINK serialization coverage #1618 added); zero-drift run at close

rubentalstra:
Closes #2919.

**The any-match adjudication now covers the fragment half.** A predicate whose leaf crosses a list-valued fragment attribute (`LOCATABLE.links`, `EVENT_CONTEXT.participations`, `PARTY_IDENTIFIED.identifiers`, `DV_TEXT.mappings`, ...) lowers existentially through the set-returning `jsonb_path_query` — one item per match in the (implicitly LATERAL) FROM of the EXISTS subquery, the predicate conjoined over the coerced item — for `=` and friends, `LIKE`, `MATCHES`, and `EXISTS` alike, in both the inline (anchor-empty) and anchored-walk shapes. `jsonb_path_query_first` remains only where it is exact: single-valued fragments and the scalar contexts (ORDER BY, MIN/MAX/SUM/AVG, path-vs-path, negative-polarity comparisons — the same recorded polarity design the anchored lowering has carried since #1448).

**Projection over a multi-valued fragment path is decided:** every match as ONE jsonb array cell (`NULLIF(jsonb_path_query_array(...), '[]')` — SQL NULL when nothing matches, so absence keeps reading as a null cell and `COUNT`/`EXISTS` semantics stay exact). QUERY master03 §Identified Paths is silent; the `// NOTE:` on `data_leaf_expr` records the decision. `SELECT c/links/target/value` over two links now serves `["ehr://x/first","ehr://x/last"]` where it served one scalar.

**`LeafPath.multi_valued` is consumed via its step flags:** the dead whole-path aggregate is deleted; the SQL layer reads `fragment_multi_valued()` (the per-step flags analysis was already computing).

**En-route discovery, filed as #2927 and half-fixed here:** path-embedded predicates outside the anchor walk were SILENTLY DROPPED — `c[openEHR-EHR-COMPOSITION.report.v1]/name/value = 'x'` built SQL with no archetype condition, and `c/links[at0001]/...` discarded the constraint. The root predicate now lowers (CASE-guarded inline/anchored reads, conjoined in existential subqueries); the two shapes with no lowering yet — a fragment-step predicate, and a root predicate on a whole-object projection — are typed rejects carrying `TODO(#2927)`, per the engine's reject-never-silently-answer rule.

**Pinned by:** five planner SQL-shape tests (existential + array-cell + scalar-stays-scalar + EXISTS per-match + root-predicate reach/rejects) and two DB suites committing a composition with two links, two participations and two composer identifiers where only the LAST element matches — for `=` (root and anchored), `LIKE`, `MATCHES`, `EXISTS`, the array-cell projection, and the no-vacuous-truth negatives. The `EXISTS` control had to probe `links`: the vendored base fixture already carries a participation.

**The CNF wire case for LINK filtering** is filed as Veredictum#178 (catalogue changes live in the instrument's own tracker) with the any-match register ground spelled out; the local suites carry the behavioural pinning meanwhile.

Docs: the AQL book page's any-match paragraph now covers list attributes and states the array-cell projection shape; changelog entries under `### Fixed`.

Suites: ferroehr 998/998, ferroehr-rest 773/773, workspace clippy/fmt/comment-style/rustdoc clean. The zero-drift conformance run is recorded below before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,21 @@ workflow refuses a tag that has no matching section here.
 
 ### Fixed
 
+- AQL predicates and projections now see every element of a list-valued
+  attribute (`links`, `context/participations`, composer `identifiers`,
+  `mappings`, ...). Previously the extraction took only the first match, so
+  `WHERE c/links/target/value = ...` missed a match on the second link and
+  `SELECT` over such a path served one value where the data held several —
+  silently. Comparisons, `LIKE`, `MATCHES` and `EXISTS` are any-match;
+  projection returns every match as one JSON array cell (`null` when nothing
+  matches).
+- An AQL node predicate written on the ROOT of an identified path
+  (`c[openEHR-EHR-COMPOSITION.report.v1]/name/value`) now constrains the
+  query — it was silently discarded, so the path answered as if the
+  predicate were not written. The two shapes that still have no lowering — a
+  predicate on a non-structure path step (`links[at0001]`) and a node
+  predicate on a whole-object projection — are refused with a typed error
+  instead of being ignored.
 - Reads serve the exact canonical bytes the commit accepted: `_type` first
   and every field at its spec-declared position — including the
   server-stamped `uid`, which now lands at its place in the document instead
--- a/app/ferroehr/src/aql/analyze.rs
+++ b/app/ferroehr/src/aql/analyze.rs
@@ -135,7 +135,6 @@ fn analyze_rm_path(
     let mut fragment: Vec<FragmentStep> = Vec::new();
     let mut in_fragment = false;
     let mut current = root_types.clone();
-    let mut multi_valued = false;
     // The candidate types of the *parent* of the final step, and the final
     // attribute name — used to coerce a `.../value` leaf under a temporal DV
     // parent to `Temporal` (its `value` is an ISO-8601 String; QUERY §Built-in
@@ -148,9 +147,6 @@ fn analyze_rm_path(
             parent_types = current.clone();
             last_name = Some(part.name.clone());
             let (step_types, is_multi) = resolve_attribute(&current, &part.name, profile)?;
-            if is_multi {
-                multi_valued = true;
-            }
             let predicate = part
                 .predicate
                 .as_ref()
@@ -195,7 +191,6 @@ fn analyze_rm_path(
         fragment,
         types: current,
         coercion,
-        multi_valued,
     })
 }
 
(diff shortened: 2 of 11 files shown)


rubentalstra:
Zero-drift conformance run on this branch: 1102 case-records, 1064 passed / 0 failed / 0 errored / 38 n-a — the committed artifacts came back byte-identical (no motion at all).

